PR #166, opened 2022-10-07
Correction on L1.CentralFreq error by changing L1.CentralFreq.data + …

ACHMartin:
…modification of the test notebook.


DavidMcCann-NOC (commented):
This should work - we might have to keep an eye out for future iterations of this problem with single numbers stored as DataArrays

PR #172, opened 2022-10-20
add two notebooks + remove crop_function in tools

ACHMartin:


DavidMcCann-NOC (commented):
These should be useful, thanks

PR #184, opened 2022-11-03
Recent changes

DavidMcCann-NOC:
- X-band .mat file reader and colocation tool
- Translation of Karlus' incidence angle script (although it turns out we may have an outdated version now, so this may need updating in the future)
- Docstring standardisation and Sphinx setup files for automatic documentation. Now working but will need some HTML style formatting


seastar/oscar/level1.py
```diff
@@ -224,10 +224,160 @@ def compute_incidence_angle(ds):
...
+            + (N - gridinfo[1]) * np.sin(gridinfo[8])
+    orb_z = ds.OrbHeightImage
+    xlen = len(x_axis)
+    ylen = len(y_axis)
+
+#speedup_factor_x = 50
+#speedup_factor_y = 50
+    speedup_factor_x = 48
+    speedup_factor_y = 40
+    orbx = xr.DataArray(data=orb_x,
```
ACHMartin, line 283: just for your information, no need to change the code, you can certainly do coarsen/rolling mean directly in numpy without the need to call for xarray. It is not useful here and it makes it more difficult to read/understand why/how xarray is used. (my understanding is it is only used for coarsen it)

seastar/oscar/level1.py
```diff
@@ -224,10 +224,160 @@ def compute_incidence_angle(ds):
     nadir (radians) for each pixel
 
     """
-    X, Y = np.meshgrid(ds.CrossRange, ds.GroundRange, indexing='ij')
-    ds['IncidenceAngleImage'] = np.degrees(np.arctan(np.sqrt(2) *
-                                                     Y / ds.OrbHeightImage))
```
ACHMartin, line 244: perhaps the old function could be renamed:
compute_incidence_angle_from_simple_geometry or something more explicit than simple_geometry

the other one might be called:
compute_incidence_angle_from_GBPGridInfo

seastar/oscar/level1.py
```diff
@@ -1,10 +1,7 @@
 # -*- coding: utf-8 -*-
 """
-Functions to calculate varirous L2 data products for the OSCAR instrument.
```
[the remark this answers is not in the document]
DavidMcCann-NOC: I had to stare at that for an inordinately long time to notice!

seastar/oscar/level1.py
```diff
@@ -327,8 +326,48 @@ def compute_radial_surface_velocity(ds):
 
     return ds
 
+def compute_radial_surface_current(level1, aux, gmf='mouche12'):
```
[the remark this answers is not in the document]
DavidMcCann-NOC: I also need to learn what **kwargs really are and how to use them properly to be honest

seastar/oscar/level1.py
```diff
@@ -224,10 +224,160 @@ def compute_incidence_angle(ds):
...
+        inc_angle = xr.DataArray(data=inc_angle,
+                                 coords=[xaxis, yaxis],
+                                 dims=('CrossRange', 'GroundRange')
+                                 )
+        look_angle = xr.DataArray(data=look_angle,
+                                  coords=[xaxis, yaxis],
+                                  dims=('CrossRange', 'GroundRange')
+                                  )
+
+    ds['IncidenceAngleImage'] = inc_angle.interp_like(ds.CrossRange)
```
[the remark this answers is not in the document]
DavidMcCann-NOC: Not sure If what i did was right or not but it comes out as we would expect (minus some edge effects from the NaN inpainting). This should be adapted to scrub the mention of CrossRange though, this was just a quick fix to get it running

seastar/oscar/level1.py
DavidMcCann-NOC: Thanks - I agree, it does make it a little more unnecessarily complicated. I like the coarsen functionality though so in the future i'll get it working directly in numpy

ACHMartin: if there is already an xarray dataset it is fine to work directly on it, just here I wonder why we had a dataset and not using it after

Issue #170, opened 2022-10-18
Attributes need writing for computed variables


DavidMcCann-NOC:
Currently attributes e.g., 'long_name', 'units' are not present for computed data arrays. Need to add attributes at the end of computation functions.

ACHMartin:
Priority for the Level2 data fields than the intermediate L1 products

DavidMcCann-NOC:
Closed with commit #188

DavidMcCann-NOC:
Long name should be just 'Current Velocity' rather than the current long names. Units without parentheses.

Revisit all dataarrays 

Look in to differences between 'long_name', 'standard_name' etc and their impact on plotting.

DavidMcCann-NOC:
Perhaps a new attribute field with the description in it?

DavidMcCann-NOC:
Attributes to be standardized in the following format:

- ``[long_name]`` = Short title for plotting purposes, e.g., ``'Radial surface current'``
- ``[description]`` = Long title, e.g., ``'Radial Surface Current (RSC) along antenna beam direction, corrected for Wind Artifact Surface Velocity (WASV)'``
- ``[units]`` = units in SI format as a ``str``, without brackets. e.g., ``'m/s'``

PR #187, opened 2022-11-07
Import inc angle and squint function

DavidMcCann-NOC:
Function to import pre-computed incidence angle and squint fields into dataset from .mat files. 

- Fields will need pre-computing for each OSCAR dataset supplied by metasensing. Currently only computed for the 17/05.

### DIFF
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -433,14 +433,47 @@ def compute_antenna_azimuth_direction(ds, antenna):
         'Heading (degrees N) of the airfraft for each pixel in the image'
     ds.OrbitHeadingImage.attrs['units'] = '[degrees]'
 
-    # Assuming antennas pointing to port. If pointing to starboard
-    # then change sign
-    if antenna == 'fore':
-        ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage - 45, 360)
-    elif antenna == 'aft':
-        ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage - 135, 360)
-    else:
-        raise 'Error unknown antennas str'
+    lookdirec = re.sub('[^LR]', '', str(ds.LookDirection.data))
+    if lookdirec == 'L':
+        if 'SquintImage' in ds:
+            if antenna == 'fore':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   - ds.SquintImage,
+                                                   360)
+            elif antenna == 'aft':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage -
+                                                   (np.abs(90 -
+                                                           ds.SquintImage)),
+                                                   360)
+        elif 'SquintImage' not in ds:
+            if antenna == 'fore':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   - 45,
+                                                   360)
+            elif antenna == 'aft':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage -
+                                                   135,
+                                                   360)
+    elif lookdirec == 'R':
+        if 'SquintImage' in ds:
+            if antenna == 'fore':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   + ds.SquintImage,
+                                                   360)
+            elif antenna == 'aft':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage +
+                                                   (90
+                                                    + np.abs(ds.SquintImage)),
+                                                   360)
+        elif 'SquintImage' not in ds:
+            if antenna == 'fore':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   + 45,
+                                                   360)
+            elif antenna == 'aft':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   + 135,
+                                                   360)
 
     return ds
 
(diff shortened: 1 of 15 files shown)


seastar/oscar/level1.py
```diff
@@ -433,14 +433,47 @@ def compute_antenna_azimuth_direction(ds, antenna):
...
+                                                   360)
+        elif 'SquintImage' not in ds:
+            if antenna == 'fore':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   + 45,
+                                                   360)
+            elif antenna == 'aft':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   + 135,
+                                                   360)
```
ACHMartin, line 476: Same here.

seastar/oscar/level1.py
```diff
@@ -433,14 +433,47 @@ def compute_antenna_azimuth_direction(ds, antenna):
...
+        elif 'SquintImage' not in ds:
+            if antenna == 'fore':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage
+                                                   - 45,
+                                                   360)
+            elif antenna == 'aft':
+                ds['AntennaAzimuthImage'] = np.mod(ds.OrbitHeadingImage -
+                                                   135,
+                                                   360)
```
ACHMartin, line 456: should we raise an error if SquintImage is not produced? I am worried it will go on unnotice and we will struggle to find the origine of the (small) error.

seastar/oscar/level1.py
```diff
@@ -489,8 +522,8 @@ def compute_radial_surface_velocity(ds):
     """
     if 'CentralWavenumber' not in ds.data_vars:
         ds = add_central_electromagnetic_wavenumber(ds)
-    
-    ds = compute_incidence_angle(ds)
+    if 'IncidenceAngleImage' not in ds:
+        ds = compute_incidence_angle_from_simple_geometry(ds)
```
ACHMartin, line 526: If the field is not defined, I think it will be safer to raise an error, otherwise I am worried it will be unnoticed what/when this calculation is done.

ACHMartin (commented):
Great, nice these small and easy pull request!

seastar/utils/tools.py
```diff
@@ -179,6 +179,21 @@ def compute_relative_wind_direction(windDirection, lookDirection):
...
+    data_vars = list(data.keys())
+    for var_name in data_vars:
+        var_data = data[var_name]
+        if isinstance(var_data, np.ndarray):
+            ds[var_name] = xr.DataArray(
+                data=np.degrees(var_data),
+                coords={'GroundRange': ds.GroundRange,
+                        'CrossRange': ds.CrossRange},
+                dims=('GroundRange', 'CrossRange')
+                ).T
```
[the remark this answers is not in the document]
DavidMcCann-NOC: Good idea - its probably best if I start getting ahead of the attributes writing as I go along, especially as its an open issue

seastar/oscar/level1.py
DavidMcCann-NOC: Yes definitely - Just starting to get used to warnings and errors / exceptions, I'll ad one in.

seastar/utils/tools.py
```diff
@@ -179,6 +179,21 @@ def compute_relative_wind_direction(windDirection, lookDirection):
         )
     return relative_wind_direction
 
+def import_incidence_angle_and_squint_to_dataset(filename,ds):
```
ACHMartin, line 182: perhaps in the description of this function, it would be very useful to have the basic processing to run the Matlab scripts/function, for example:

script/function works with Matlab version XXX
need to modify script xxx.m
run with
matlab directory_xxx/xxx.m
it will save the file xxx.mat in the directory_yyy

Potentially, you could add all these files in the SeaSTAR GitHub project.

DavidMcCann-NOC: I was wondering about matlab files on the repo - this is a really good idea, not least to keep everything in one place

seastar/oscar/level1.py
DavidMcCann-NOC: I'll raise a warning for now, so as not to interrupt the computation flow, but I'll make it sufficiently transparent so we know whats happened

seastar/utils/tools.py
DavidMcCann-NOC: This function is obsolete now as the missing fields are / can be pre-computed using a matlab script on the github that adds the fields to the netcdf file

seastar/utils/tools.py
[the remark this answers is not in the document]
DavidMcCann-NOC: I'm going to subsume this into issue #170 and add all of the missing attributes in one go

seastar/utils/tools.py
ACHMartin: I agree, you can add this in the header, but probably good to keep this function alive in case we change our mind concerning how to post-process the data.

seastar/oscar/level1.py
```diff
@@ -221,18 +220,34 @@ def compute_local_coordinates(ds):
...
 def compute_incidence_angle_from_simple_geometry(ds):
 
     X, Y = np.meshgrid(ds.CrossRange, ds.GroundRange, indexing='ij')
-    ds['IncidenceAngleImage'] = np.degrees(np.arctan(np.sqrt(2) *
-                                                     Y / ds.OrbHeightImage))
+    if 'SquintImage' in ds:
+        ds['IncidenceAngleImage'] = np.degrees(
+            np.arctan(
+                np.abs(ds.SquintImage)
+                * Y
```
ACHMartin, line 234: it should be
Y / cosd(ds.SquintImage) / ds.OrbHeightImage

DavidMcCann-NOC: Ah, good catch!

Issue #186, opened 2022-11-04
Error in retrieved current direction


DavidMcCann-NOC:
There are major errors present in the retrieved current direction, namely in compute_current_magnitude_and_direction. When compared to MARS2D surface currents there is very poor agreement, but alarmingly also a lack of data points in particular current directions:

![image](https://user-images.githubusercontent.com/107252303/200002975-2b2e3c0d-1adb-47e9-9599-a1b6b3944246.png)

![image](https://user-images.githubusercontent.com/107252303/200003012-cd0b703f-5c28-4381-b0e1-675ac2820c8f.png)

DavidMcCann-NOC:
This seems to be resolved over a number of commits. U,V vectors match those in the x_band data using the same functions to calculate direction

PR #188, opened 2022-11-09
New current direction retireval code and attribute writing

DavidMcCann-NOC:
New current retrieval code to address issue #186 . New code uses slightly more transparent trig and atan rather than acos, effectively just calculating vector components for each beam, then finding the inverse tangent of their combination. Tested at first on the X-band current vectors to re-create the directions Ruben sent us in his .kmz file.

The results didn't look good at first but there was an additional bug (on my end) with an erroneous wind direction. With this corrected the WASV is better and the comparison is now pretty good:
image
image

### DIFF
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -78,6 +78,9 @@ def check_antenna_polarization(ds):
     polarization = [str(ds.TxPolarization.data), str(ds.RxPolarization.data)]
     polarization = ''.join(polarization)
     ds['Polarization'] = re.sub('[^VH]', '', polarization)
+    ds.Polarization.attrs['long_name'] =\
+        'Antenna polarization, Transmit / Recieve.'
+    ds.Polarization.attrs['units'] = '[none]'
     return ds
 
 
@@ -89,7 +92,7 @@ def add_antenna_baseline(ds, baseline):
     ----------
     ds : ``xarray.Dataset``
         OSCAR SAR dataset
-    baseline : ``float``
+iat    baseline : ``float``
         Antenna baseline (m)
 
     Returns
@@ -102,6 +105,9 @@ def add_antenna_baseline(ds, baseline):
     """
     if 'Baseline' not in ds.data_vars:
         ds['Baseline'] = baseline
+        ds.Baseline.attrs['long_name'] =\
+            'Antenna ATI baseline'
+        ds.Baseline.attrs['units'] = '[m]'
     return ds
 
 
@@ -126,9 +132,15 @@ def compute_SLC_Master_Slave(ds):
     """
     ds['SigmaSLCMaster'] = (ds.SigmaImageSingleLookRealPart + 1j *
                             ds.SigmaImageSingleLookImaginaryPart)
+    ds.SigmaSLCMaster.attrs['long_name'] =\
+        'Single Look Complex image, Master antenna'
+    ds.SigmaSLCMaster.attrs['units'] = '[radians]'
     if 'SigmaImageSingleLookRealPartSlave' in ds.data_vars:
         ds['SigmaSLCSlave'] = (ds.SigmaImageSingleLookRealPartSlave + 1j *
                                ds.SigmaImageSingleLookImaginaryPartSlave)
+        ds.SigmaSLCSlave.attrs['long_name'] =\
+            'Single Look Complex image, Slave antenna'
+        ds.SigmaSLCSlave.attrs['units'] = '[radians]'
     return ds
 
 
@@ -152,6 +164,9 @@ def add_central_electromagnetic_wavenumber(ds):
 
     """
     ds['CentralWavenumber'] = 2 * np.pi * ds.CentralFreq / sp.constants.c
+    ds.CentralWavenumber.attrs['long_name'] =\
+        'Central electromagnetic wavenumber'
+    ds.CentralWavenumber.attrs['units'] = '[radians / m]'
     return ds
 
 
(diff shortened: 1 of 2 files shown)


DavidMcCann-NOC:
It may well be that remaining errors are to do with the WASV and their impact on the RSC. I'm pretty confident that the direction retrieval trig is good as it reproduces test data correctly. But we can re-assess this in the future when we have a better handle on the winds perhaps?

seastar/oscar/level1.py
```diff
@@ -89,7 +92,7 @@ def add_antenna_baseline(ds, baseline):
     ----------
     ds : ``xarray.Dataset``
         OSCAR SAR dataset
-    baseline : ``float``
+iat    baseline : ``float``
```
ACHMartin, line 95: bug with the "iat"?

seastar/oscar/level1.py
```diff
@@ -78,6 +78,9 @@ def check_antenna_polarization(ds):
     polarization = [str(ds.TxPolarization.data), str(ds.RxPolarization.data)]
     polarization = ''.join(polarization)
     ds['Polarization'] = re.sub('[^VH]', '', polarization)
+    ds.Polarization.attrs['long_name'] =\
+        'Antenna polarization, Transmit / Recieve.'
```
ACHMartin, line 82: Typo on receive

seastar/oscar/level1.py
```diff
@@ -126,9 +132,15 @@ def compute_SLC_Master_Slave(ds):
     """
     ds['SigmaSLCMaster'] = (ds.SigmaImageSingleLookRealPart + 1j *
                             ds.SigmaImageSingleLookImaginaryPart)
+    ds.SigmaSLCMaster.attrs['long_name'] =\
+        'Single Look Complex image, Master antenna'
+    ds.SigmaSLCMaster.attrs['units'] = '[radians]'
```
ACHMartin, line 137: It is a ratio of power with some phase information. So there is no units, but in practice people tend to say [linear] or [dB]. I think it is probably better to put [None] at this stage

seastar/oscar/level1.py
```diff
@@ -126,9 +132,15 @@ def compute_SLC_Master_Slave(ds):
...
     if 'SigmaImageSingleLookRealPartSlave' in ds.data_vars:
         ds['SigmaSLCSlave'] = (ds.SigmaImageSingleLookRealPartSlave + 1j *
                                ds.SigmaImageSingleLookImaginaryPartSlave)
+        ds.SigmaSLCSlave.attrs['long_name'] =\
+            'Single Look Complex image, Slave antenna'
+        ds.SigmaSLCSlave.attrs['units'] = '[radians]'
```
ACHMartin, line 143: same here

seastar/oscar/level1.py
```diff
@@ -187,17 +202,29 @@ def compute_multilooking_Master_Slave(ds, window=3):
         ds = compute_SLC_Master_Slave(ds)
     ds['IntensityAvgComplexMasterSlave'] = (ds.SigmaSLCMaster * np.conjugate(ds.SigmaSLCSlave))\
         .rolling(GroundRange=window).mean().rolling(CrossRange=window).mean()
+    ds.IntensityAvgComplexMasterSlave.attrs['long_name'] =\
+        'Average intensity of Master/Slave single looc complex images'
+    ds.IntensityAvgComplexMasterSlave.attrs['units'] = '[radians]'
```
ACHMartin, line 207: change to None

seastar/oscar/level1.py
```diff
@@ -187,17 +202,29 @@ def compute_multilooking_Master_Slave(ds, window=3):
...
     ds['Intensity'] = np.abs(ds.IntensityAvgComplexMasterSlave)
+    ds.Intensity.attrs['long_name'] =\
+        'Absolute single look complex image intensity'
+    ds.Intensity.attrs['units'] = '[radians]'
```
ACHMartin, line 211: change to None

seastar/oscar/level1.py
```diff
@@ -187,17 +202,29 @@ def compute_multilooking_Master_Slave(ds, window=3):
...
     ds['Interferogram'] = (
         ['CrossRange', 'GroundRange'],
         np.angle(ds.IntensityAvgComplexMasterSlave, deg=False)
         )
+    ds.Interferogram.attrs['long_name'] =\
+        'Interferogram between master/slave antenna pair'
+    ds.Interferogram.attrs['units'] = '[radians]'
```
ACHMartin, line 218: Radians is correct

seastar/oscar/level1.py
```diff
@@ -187,17 +202,29 @@ def compute_multilooking_Master_Slave(ds, window=3):
...
     ds['IntensityAvgMaster'] = (np.abs(ds.SigmaSLCMaster) ** 2)\
         .rolling(GroundRange=window).mean().rolling(CrossRange=window).mean()
     ds['IntensityAvgSlave'] = (np.abs(ds.SigmaSLCSlave) ** 2)\
         .rolling(GroundRange=window).mean().rolling(CrossRange=window).mean()
     ds['Coherence'] = ds.Intensity / np.sqrt(ds.IntensityAvgMaster
                                              * ds.IntensityAvgSlave)
+    ds.Coherence.attrs['long_name'] =\
+        'Coherence between master/slave antenna pair'
+    ds.Coherence.attrs['units'] = '[]'
```
ACHMartin, line 227: What is the best, to put [] or [None]?
You put [none] for polarisation. Is there any recommandation?

ACHMartin:
I am happy to discuss the points on the name convention for the geometry/projection if needed.

seastar/retrieval/level2.py
```diff
@@ -35,31 +35,63 @@ def compute_current_magnitude_and_direction(level1, level2):
...
+    u_1 = level1.sel(Antenna='Fore').RadialSurfaceCurrent\
+        * np.cos(np.radians(level1.sel(Antenna='Fore').AntennaAzimuthImage))
+    v_1 = level1.sel(Antenna='Fore').RadialSurfaceCurrent\
+        * np.sin(np.radians(level1.sel(Antenna='Fore').AntennaAzimuthImage))
+    u_2 = level1.sel(Antenna='Aft').RadialSurfaceCurrent\
+        * np.cos(np.radians(level1.sel(Antenna='Aft').AntennaAzimuthImage))
+    v_2 = level1.sel(Antenna='Aft').RadialSurfaceCurrent\
+        * np.sin(np.radians(level1.sel(Antenna='Aft').AntennaAzimuthImage))
+
+    direction = np.degrees(np.arctan2((u_1 + u_2), (v_1 + v_2)))
```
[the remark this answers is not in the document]
DavidMcCann-NOC: This all makes sense - I think what I did is swapped the order in arctan2 to compensate for getting the sin and cos the wrong way round earlier.

Agreed about the 90 degrees - I may have ended up doing that further downstream when plotting and had yet to put it in to the function. There's always some conversion needed isn't there!

seastar/oscar/level1.py
DavidMcCann-NOC: I'm not sure - something to check. In our Oscar datasets there are unitless variables with '[]' as their units. I think 'none' is more transparent

ACHMartin: In https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03.html
for all the ratio, the unit should be [1].
For the coherence, it is also a ratio, so should be [1] as well

ACHMartin: We can find the list of unit there:
following https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2.html#Database
For radian, it is singular unit [rad] or [radian]

PR #192, opened 2022-11-16
Functions to improve data import

DavidMcCann-NOC:
New small functions to:
- find_file_triplets, uses list_duplicates to find OSCAR .nc files with the same aquisition time (from filename, currently) and combine the three beam files as a triplet of indices
- antenna_ident uses identify_antenna_location to use the processed doppler to find which beam each dataset belongs to
- load_OSCAR_data to read in the triplet of .nc files into a dataset dict
- Changes to merge_beams to merge the dict of datasets, assigning Antenna dimensions based on antenna_idents
- fill_missing_variables to get around missing variables between datasets. This is a running repair and should be obsolete when we find a better way of doing it within xarray.


seastar/oscar/level1.py
```diff
@@ -12,7 +12,55 @@
 import warnings
 
 
-def merge_beams(dsf, dsa, dsm):
+def fill_missing_variables(ds, antenna_ident):
```
ACHMartin, line 15: I have the impression the name does not reflect what the function is doing.
It doesn't fill, it lists/identify the missing variables between each other

seastar/oscar/level1.py
```diff
@@ -12,7 +12,55 @@
...
+    ds_diff.where(ds_diff == np.nan, other=np.nan)
+    ds[antenna_ident.index('Fore')] = ds[antenna_ident.index('Fore')]\
+        .merge(ds_diff)
+
+    ds_diff = ds[antenna_ident.index('Mid')]\
+        [[x for x in ds[antenna_ident.index('Mid')].data_vars
+          if x not in ds[antenna_ident.index('Aft')].data_vars]]
+    ds_diff.where(ds_diff == np.nan, other=np.nan)
+    ds[antenna_ident.index('Aft')] = ds[antenna_ident.index('Aft')]\
+        .merge(ds_diff)
```
ACHMartin: I think you can do a loop with pairs [['Fore', 'Mid'], ['Mid', 'Fore'], ... (cf what we have with pol_val, pol_str)
it will make think easier to read.

seastar/oscar/level1.py
```diff
@@ -21,37 +69,47 @@ def merge_beams(dsf, dsa, dsm):
...
+        OSCAR data stored as a dict with antenna number as keys and loaded
+        data in ``xarray.Dataset`` format as values
+    antenna_ident : ``list``
+        List of antenna identifiers in the form ['Fore', 'Mid', 'Aft'],
+        corresponding to the data and keys stored in `ds`
 
     Returns
     -------
-    ds_level1 : ``xarray.Dataset``
+    dsl1 : ``xarray.Dataset``
```
ACHMartin, line 82: probably better to keep expansion of l1 as level1 cf PEP8, l-L can be misinterpret as a i-I

seastar/oscar/level1.py
```diff
@@ -21,37 +69,47 @@ def merge_beams(dsf, dsa, dsm):
...
         OSCAR dataset with combined look directions
 
     """
+    dsl1 = xr.concat(list(ds.values()),
+                     'Antenna', join='inner',
```
ACHMartin: Should we keep inner or move to outer?

seastar/oscar/level1.py
```diff
@@ -12,7 +12,55 @@
...
+        [[x for x in ds[antenna_ident.index('Mid')].data_vars
+          if x not in ds[antenna_ident.index('Aft')].data_vars]]
+    ds_diff.where(ds_diff == np.nan, other=np.nan)
+    ds[antenna_ident.index('Aft')] = ds[antenna_ident.index('Aft')]\
+        .merge(ds_diff)
+
+    return ds
+
+
+def merge_beams(ds, antenna_ident):
```
ACHMartin, line 64: 'ds' is the typical name for a xarray DataSet, here we have a dictionary of ds. I don't remember what is often used for a dictionary, perhaps we could use dict_ds.

Do we need antenna_ident or can we get it directly from dict_ds, e.g. dict_ds.keys()

ACHMartin, line 64: If we keep "antenna_ident", I find it difficult to understand what it is. I tend to read "indent" instead of "ident".
I would prefer "antenna_id" or "antenna_id_list", "antenna_keys", ...

seastar/oscar/level1.py
ACHMartin, line 15: Should fill_missing_variable be a sub function of merge_beam (so within the definition of merge_beam), and to be called within the function.

ACHMartin, line 15: see comments in merge_beams concerning ds->ds_dict and antenna_ident

seastar/oscar/level1.py
DavidMcCann-NOC: I think to do this I'd have to re-think how I do the comparison and merging. I agree its a little ham-fisted though... I'll add some comments and try and space it out for now to make it a bit more understandable.

The function is just a way to get around variables that appear in one dataset but not another. I think a cleaner way would be to loop through the antennas, build a combined var_list and then do the merging / NaN filling with that. Might make this when I've got some time

seastar/oscar/level1.py
DavidMcCann-NOC: Good point

seastar/oscar/level1.py
DavidMcCann-NOC: Strange, reads as 'outer' on my version. Either way I'll merge this file in - we should be able to stick with 'outer' from now on

seastar/oscar/level1.py
DavidMcCann-NOC: Good point, i think ds_dict is the better name as its a dict of datasets, although either way it says the same thing.

antenna_ident is built using a function that interrogates the doppler data from each dataset and is agnostic to how the files are loaded in. We could technically do this inside functions but its also very useful outside them for passing antenna identity to other computations, plotting names etc. To do this internally we would call antenna_idents again like:

antenna_idents = seastar.utils.tools.antenna_idents(ds)

DavidMcCann-NOC: Thats fair - antenna_id is a good idea

seastar/oscar/level1.py
ACHMartin: I agree, probably good to open an issue and link it to here somehow with a detail idea on how you want to proceed. A "nice to have" flag will probably be good enough

seastar/oscar/level1.py
ACHMartin: In my opinion, if we can get the informations internally using the function, there is no need to double input

PR #196, opened 2022-11-18
Adrien nscat4 + mouche12, yuro19 adaptation for XR.DataArray

ACHMartin:
- Addition of NSCAT4DS GMF for NRCS
- modification of mouche12 and yuro19 to get XR.DataArray as outputs
- Addition of notebooks with plots as function of incidence angle, wspd and wdir


DavidMcCann-NOC (approved):
Not much to review on my end, all looks good. I look forward to see it all in action!

PR #200, opened 2022-11-22
Changes to get star-pattern runs working in the current build

DavidMcCann-NOC:
Commits to get the May 22nd data working in the current build


ACHMartin (commented):
all good

Issue #206, opened 2022-12-02
Enabling multi-dimension xr.Dataset in compute_wasv


ACHMartin:
Currently compute_wasv doesn't allow an Antenna dimension. It will help if it can allow it. Perhaps move some elements as alignment outside the function. Keep some form of test if the data are correctly aligned. + update the documentation on the alignment.

DavidMcCann-NOC:
Closed with #211

Issue #204, opened 2022-12-02
LookDirection or AntennaAzimuthImage


ACHMartin:
In Doppler (perhaps other section) is LookDirection or AntennaAzimuthImage we should be used?

ACHMartin:
Using "LookDirection" in "test_seastar_doppler_package_adrien_v0.ipynb" it raises an error.
We need to correct all files + documentation where "LookDirection" is mentioned, including modification of the notebook. It is a standalone notebook without the local path, so I think we could remove the "adrien".

DavidMcCann-NOC:
I agree its confusing semantics. LookDirection as the variable Metasensing have provided denotes the port/starboard pointing direction of the arrays and is a single letter L or R (always L in our case). So I agree, perhaps we need to find a different phrase for the documentation in other functions. Perhaps 'Antenna direction' or 'Antenna location'?

The only place we actually use it in Doppler is the exception:
if len(L1.LookDirection.shape) > 2:
        raise Exception('L1.LookDirection need to be a 2D field. \n'
                        'Use e.g. L1.sel(Antenna="Fore") to reduce to '
                        'a 2D field.')
Which looks to me to be expecting IncidenceAngleImage?

ACHMartin:
Thank you for the explanation. Indeed, either IncidenceAngleImage or AntennaAzimuthImage.

ACHMartin:
link to #211

DavidMcCann-NOC:
Closed with #211

PR #211, opened 2022-12-06
Multi-beam implementation of compute_wasv

DavidMcCann-NOC:
Changes to compute_wasv and compute_radial_surface_current to accept the full beam-merged level1 dataset. Not currently tested for yurovsky but identical results obtained for mouche as previous implementation.

### DIFF
--- a/seastar/gmfs/doppler.py
+++ b/seastar/gmfs/doppler.py
@@ -9,7 +9,7 @@ from scipy.constants import c  # speed of light in vacuum
 import seastar
 
 
-def compute_wasv(L1, aux_geo, gmf, **kwargs):
+def compute_wasv(L1_combined, aux_geo, gmf, **kwargs):
     """
     Compute the Wind-wave Artefact Surface Velocity (WASV).
 
@@ -20,7 +20,7 @@ def compute_wasv(L1, aux_geo, gmf, **kwargs):
     ----------
     L1 : xarray.Dataset
         A Dataset containing IncidenceAngleImage, LookDirection and
-        Polarization.
+        Polarization across three beam directions ('Fore', 'Aft', 'Mid')
     aux_geo : xarray.Dataset
         A Dataset containing WindSpeed, WindDirection.
     gmf : str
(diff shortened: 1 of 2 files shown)


DavidMcCann-NOC:
To address issue #206

ACHMartin:
I didn't think about this yesterday for compute_nrcs, but as it is implemented with this new version, we can't use compute_nrcs or compute_wasv if we don't have an Antenna dimension.
After discussion: 
1st step/solution, to put a dimension string as input of the function (in our case it will be Antenna), how does it work if we only have a 2D field?
2nd step/solution, to have a function agnostic of dimension. I think the main issue is from xr.align(L1,geo), perhaps we can just test the data are aligned and raised issue if it is not the case, but not aligning within the function.

ACHMartin:
> Perhaps a good way to implement this would be to have the wasv and nrcs computations work on DataArrays, with whatever function that calls them (e.g. compute_radial_surface_current) do the dimension wrangling. I've been working on this for a couple of hours and can make the dimension compute_wasv runs over dimension-agnostic but can't get it to do both that and work on just data arrays (0th dimension, essentially) without an if and repeating the code all over again for the different case.

In this case, perhaps the code you need to repeat can be a sub-function. We might discuss the code together on Friday if you wish

ACHMartin:
what is the status of this? Are you ready to deploy it? Thanks

DavidMcCann-NOC:
> what is the status of this? Are you ready to deploy it? Thanks

I think its good to merge, it works in the current 'Antenna' environment we're operating in but will need some more work to make it fully dimension agnostic

ACHMartin:
> Perhaps a good way to implement this would be to have the wasv and nrcs computations work on DataArrays, with whatever function that calls them (e.g. compute_radial_surface_current) do the dimension wrangling. I've been working on this for a couple of hours and can make the dimension compute_wasv runs over dimension-agnostic but can't get it to do both that and work on just data arrays (0th dimension, essentially) without an if and repeating the code all over again for the different case.

The code as it is, is not dimension-agnostic. Antenna is mentioned. Did you forget to push a commit?

ACHMartin:
Could you please solve the LookDirection issue #204 with this merge request? Thanks

DavidMcCann-NOC:
> > Perhaps a good way to implement this would be to have the wasv and nrcs computations work on DataArrays, with whatever function that calls them (e.g. compute_radial_surface_current) do the dimension wrangling. I've been working on this for a couple of hours and can make the dimension compute_wasv runs over dimension-agnostic but can't get it to do both that and work on just data arrays (0th dimension, essentially) without an if and repeating the code all over again for the different case.
> 
> The code as it is, is not dimension-agnostic. Antenna is mentioned. Did you forget to push a commit?

I'm either living in some sort of fever dream or I've somehow managed to write a day's worth of code, fail to commit it and then loose it to the ether. You're right - its not dimension agnostic, but the code I appear to have lost was....

ACHMartin:
> I'm either living in some sort of fever dream or I've somehow managed to write a day's worth of code, fail to commit it and then loose it to the ether. You're right - its not dimension agnostic, but the code I appear to have lost was....
Could it be on a different branch?

DavidMcCann-NOC:
> > I'm either living in some sort of fever dream or I've somehow managed to write a day's worth of code, fail to commit it and then loose it to the ether. You're right - its not dimension agnostic, but the code I appear to have lost was....
> 
> Could it be on a different branch?

I've only been working on two and I've checked both. Unfortunately it seems to have fallen through the cracks which is incredibly frustrating. Thankfully I remember the structure of what i implemented so it shouldn't take too long to re-do, but perhaps when we talk next we will think up a better way of implementing this anyway

PR #214, opened 2022-12-12
Adrien cost function minimisation

ACHMartin:
the commits are relatively independent, so it should be easier to proofread by commit instead than by files.


DavidMcCann-NOC (commented):
All good, should come in handy

DavidMcCann-NOC (commented):
The dotdict class is very interesting - I hadn't thought about changing things like that

DavidMcCann-NOC:
Happy to merge this as is - all looks good to me. I'll see how I get on with the new functions

PR #218, opened 2022-12-15
Error in Aft beam direction convention fix

DavidMcCann-NOC:
Changing the direction convention of the Aft beam to fix the error in retrieved currents. Additionally re-inserting the 90 degree addition to the retrieved current
direction as this had been erroneously removed in the past.


DavidMcCann-NOC:
Commented-out the changes so I have a reference for local application until we find the physical source of this error

PR #219, opened 2022-12-17
Adrien star pattern

ACHMartin:
Two new notebooks:
- Add the analysis on the star pattern with RVL, Sigma0, Coherence as function of azimuth at different ground range
- test of the cost_function


ACHMartin:
@DavidMcCann-NOC, I don't think you will have any comment. I let you merge on your own time.

DavidMcCann-NOC (approved):
All looks good

PR #220, opened 2023-01-12
Small tool to generate a land mask for Ushant island

DavidMcCann-NOC:


seastar/utils/tools.py
```diff
@@ -505,4 +507,39 @@ class dotdict(dict):
...
+    
+    Generates a boolean land mask using the GSHHS coasline dataset. A good
+    value for `bound` to cover the island of Ushant is:
+        [-5.2, -5.0, 48.37, 48.5]
+
+    Parameters
+    ----------
+    dsl2 : ``xarray.Dataset``
+        Dataset containing 
+    bound : ``list``
```
ACHMartin, line 525: perhaps we can put "boundary" as an option using **kwargs (cf doppler/compute_total_surface_motion for example). An other possibility, is to set up the default "boundary=None" in the header and we test it if it is not None. Perhaps easier.
The default behaviour could be to use the Oscar fields Latitude11, 12, 21, 22 and take the min and max of all these numbers to set up the boundary.

seastar/utils/tools.py
```diff
@@ -505,4 +507,39 @@ class dotdict(dict):
...
+        DESCRIPTION.
+
+    Returns
+    -------
+    mask : ``array``, ``bool``
+        Boolean mask of Ushant island in data coordinates, xr.DataArray format.
+
+    """
+    coast = cfeature.GSHHSFeature(scale='full').intersecting_geometries(bound)
+    coast_polygon = next(coast)
```
ACHMartin, line 535: how does "next" work? Does it start from the biggest land and then go down in size?
if it behaves like this (hopefully), perhaps we can add a loop on this as function of a iterative function we can put in the header: "nb_independent_flag=1", or "nb_flagged_island", I am not particularly happy with these names...

I see something like this
`while it < nb_independent_flag:`
`    it+=1`
`   coast_polygon = next(coast)`

Does it work for both Level1, Level2 OSCAR data?

seastar/utils/tools.py
```diff
@@ -505,4 +507,39 @@ class dotdict(dict):
...
+    """
+    Compute land mask.
+    
+    Generates a boolean land mask using the GSHHS coasline dataset. A good
+    value for `bound` to cover the island of Ushant is:
+        [-5.2, -5.0, 48.37, 48.5]
+
+    Parameters
+    ----------
+    dsl2 : ``xarray.Dataset``
```
ACHMartin, line 523: change to "ds", we don't want let think it is only for level2 data

seastar/utils/tools.py
DavidMcCann-NOC: It will definitely need some form of boundary I think, otherwise `coast` will be generated from the entire global dataset at full resolution...

Your second point is the good way to go I think, with boundary as a **kwarg and reverting to the boundary of the data if not present. 

I'll have to test this as my gut feeling is the .intersecting_geometries() function will return the polygons that sit within the supplied boundary - i.e. it may just provide nothing unless the boundary contains the whole island. I'll check though

seastar/utils/tools.py
DavidMcCann-NOC: I think you might be right - ``coast`` is an iterative generator and i'm not 100% on how they function yet (i.e. how to find out whether we've run out of iterations). I'd like to run it to include all of the little islands - currently the first iteration (i.e. using next() is the main island coastline which is why I used it as a first case

seastar/utils/tools.py
ACHMartin: I agree, we need a boundary. The question is more in term of implementation.
We can have:
```
create_land_mask(ds, boundary=None, nb_islands=1):
   if boundary is None:
      boundary = something calculated from the data itself or from Latitude11, 12, ... field at level1
```

```
create_land_mask(ds):
   if boundary not in wards.keys():
      boundary = something calculated from the data itself or from Latitude11, 12, ... field at level1
```

ACHMartin: Ok, I understand your second point now. Sorry, to quick to read. Indeed good point to test.

DavidMcCann-NOC: Just tested it with the bounds set automatically using the latitude and longitude min/max and it works perfectly

seastar/utils/tools.py
DavidMcCann-NOC: Looks like the method to use is the `enumerate()` function

ACHMartin: so you will load all the land feature?

In practice, I think it is good if we could limit the number of feature we want. Only the 1 biggest. The reason is, if we wish to calibrate the phase, we want a big enough portion of land where we can assume it is relatively uniform. We don't want all the small islands.

DavidMcCann-NOC:
Made a few updates so it makes a little more sense to run, including the ability to use the level1 dataset as input and some better error handling of the coastline_selection. It now outputs the identified coastlines and their keys which is handy

PR #223, opened 2023-01-17
Simple config.ini reader to set local paths for each machine

DavidMcCann-NOC:
This should work on multiple devices with different operating systems, provided that the path strings for each device are entered into the config.ini file in the correct local format. (e.g., for Windows the slash used in file paths is the other way round to linux). The device name can be found with the ``platform`` module and is the output of ``platform.node()``


ACHMartin (approved):
This was quick.
Windows use "\" between folder whereas linux use "/". I understand we can get easily the path node for the OSCAR data and/or auxiliary data. How does it works if we want to navigate from these paths.
For exemple what I am doing is:
```
version = 'v20221104'

oscar_path = "../DATA/OSCAR/" + version + "/matlab-pre-processed/"

file_time_triplets = ss.utils.tools.find_file_triplets(oscar_path)
```
Is there a python solution to add path which are "platform" independent?
I see something like below for my code above:
```
version_path = add_path_with_slash(oscar_path, version)
matlab_pre_processed_path = add_path_with_slash(version_path, 'Matlab-pre-processed')
```

Issue #221, opened 2023-01-12
Data variables for the "mid" antenna which have been created (Slave, ...) in order to merge the beams are not set to NaN


ACHMartin:
For the "mid" antenna, we expect the Interferogram to be constantly NaN.

Using the code below (inspired from OSCAR_star_pattern_v1.ipynb notebooks), the interferogram is not a constant NaN but only noisy
`dsl1['Track_13'].sel(Antenna='Mid').Interferogram.plot()`
![download](https://user-images.githubusercontent.com/18679148/212058328-6babdaef-6955-42b5-8d61-6e5ff2d06089.png)

Looking further, we have:
`dsl1['Track_13'].sel(Antenna='Mid').SigmaImageSingleLookRealPartSlave.sel(CrossRange=0)` which has proper values and is not set to NaN as expected.

I assume the issue is in level1/fill_missing_variables
particularly line 46 ` ds_diff.where(ds_diff == np.nan, other=np.nan)`
In my mind `.where` does not change anything to the dataset. So, in my view it does nothing and we just copy the field from "Fore"

After correcting this, need to check the Interferogram goes to NaN as well and that xarray is not playing tricks.

On a side notes some data are set to -999. We need to put it to NaN. Probably using a new function.

```
# Data load and pre-processing
version = 'v20221104'

oscar_path = "../DATA/OSCAR/" + version + "/matlab-pre-processed/"

file_time_triplets = ss.utils.tools.find_file_triplets(oscar_path)

print('OSCAR L1 processing - version ' + version)
print('-------------------')
print('File path :', oscar_path)
print('-------------------')
for i in range(len(file_time_triplets)):
    print('File index :', i, ',', 'Aquisition time :', file_time_triplets[i][0], ',', 'File list indices :', file_time_triplets[i][1])

# Match tracks in the star pattern from track names and times in the DAR_v6.0 to file aquisition times
star_pattern_tracks={
    'Track_L1' : 0,
    'Track_1a' : 1,
    'Track_2a' : 2,
    'Track_1b' : 3,
    'Track_11' : 4,
    'Track_12' : 5,
    'Track_13' : 6,
    'Track_14' : 7,
    'Track_15' : 8,
    'Track_16' : 9,
    'Track_17' : 10,
    'Track_18' : 11}

# Compute L1C - Sigma0 + RVL
dsl1 = dict()
for track in star_pattern_tracks.keys(): # Loop through star pattern tracks
    file_index = star_pattern_tracks[track]
[...]

Issue #182, opened 2022-10-21
What outputs for each function: Full DataSet with the additional fields or the computed DataArray


ACHMartin:
We tend to do return the full DataSet with the additional field.
Should we return the DataArray instead. I think it believe of the cases, perhaps we can discuss here the cases where DataArray output should be prefer.
The new API would be:
`dsl2['newField'] = compute_new_field(dsl2)`
instead of the current:
`dsl2 = compute_new_field(dsl2)`

DavidMcCann-NOC:
I've had similar thoughts, and it wouldn't be too much effort to change. It'll be more transparent in a lot of cases I think

DavidMcCann-NOC:
Closed with #224

PR #224, opened 2023-01-18
Update to fill_missing_variables

DavidMcCann-NOC:
Here's my attempt at making sure that the mid antenna variables are properly filled with NaNs to address issue #221


seastar/oscar/level1.py
```diff
@@ -40,12 +40,19 @@ def fill_missing_variables(ds_dict, antenna_id):
...
+    for var in ds_dict[fore_id].data_vars:
+        if var not in ds_dict[mid_id].data_vars:
+            if 'Slave' in var:
+                var_master = var.replace('Slave', '')
+                ds_dict[mid_id][var] = xr.DataArray(
+                    data=np.full(ds_dict[mid_id][var_master].shape,
+                                 np.NaN),
+                    coords=ds_dict[mid_id][var_master].coords,
+                    dims=ds_dict[mid_id][var_master].dims
+                    )
```
ACHMartin, line 55: very good. Did you test how the Interferogram, Intensity and Coherence behave. Do we get a NaN or something for these?

seastar/oscar/level1.py
```diff
@@ -40,12 +40,19 @@ def fill_missing_variables(ds_dict, antenna_id):
...
     # Find vars that dont exist in Fore, but exist in Mid
     ds_diff = ds_dict[mid_id]\
```
ACHMartin, line 57: Do we continue to need to have the following lines below?
Do we have variables which are in the Mid but not in the Aft and Fore?

Anyway, the `.where` below is not doing anything and so the code is not behaving as we want.

ACHMartin (commented):
The modification you added looks great. Do we still need the hold portion of code which is faulty?
Did you try to solve the -999 issue?

seastar/oscar/level1.py
DavidMcCann-NOC: We get Nan for the Interferogram and Intensity, not checked coherence yet but it should be the same. If we want an intensity value we'll have to adapt the code to just do the square of the master for the mid antenna i guess?

DavidMcCann-NOC:
> The modification you added looks great. Do we still need the hold portion of code which is faulty? Did you try to solve the -999 issue?

I've kept the other two bits of merging code to remain for the moment as they pull in ``CalImage`` from the Mid dataset into the other two - without that they merge_beams code will still crash. Alternatively we can drop the ``CalImage`` variable upstream

ACHMartin:
> I did try and solve the -9999 issue, I have a piece of code that works nicely but strangely it gets stuck on a variable called `Dummy` that is actually the identifier for those -9999 values. In this case we have a 1x3 array of -9999 (along the antenna dimension), and when I try and replace the numbers with NaN using the same code that works for all the other arrays it complains about turning a NaN into an int.... I'll have it sorted soon i think

This sounds a very good surprise. So if I understand correctly, you only need to change this field to NaN to remove all the -999? That would be handy

ACHMartin:
> > The modification you added looks great. Do we still need the hold portion of code which is faulty? Did you try to solve the -999 issue?
> 
> I've kept the other two bits of merging code to remain for the moment as they pull in `CalImage` from the Mid dataset into the other two - without that they merge_beams code will still crash. Alternatively we can drop the `CalImage` variable upstream

What do we have in this field? Is it constant? Does it change with Tracks?

DavidMcCann-NOC:
> > > The modification you added looks great. Do we still need the hold portion of code which is faulty? Did you try to solve the -999 issue?
> > 
> > 
> > I've kept the other two bits of merging code to remain for the moment as they pull in `CalImage` from the Mid dataset into the other two - without that they merge_beams code will still crash. Alternatively we can drop the `CalImage` variable upstream
> 
> What do we have in this field? Is it constant? Does it change with Tracks?

Its the calibration constants I think - I remember Guiseppe asking about it in that meeting a while ago. I haven't looked at it fully yet

DavidMcCann-NOC:
> > I did try and solve the -9999 issue, I have a piece of code that works nicely but strangely it gets stuck on a variable called `Dummy` that is actually the identifier for those -9999 values. In this case we have a 1x3 array of -9999 (along the antenna dimension), and when I try and replace the numbers with NaN using the same code that works for all the other arrays it complains about turning a NaN into an int.... I'll have it sorted soon i think
> 
> This sounds a very good surprise. So if I understand correctly, you only need to change this field to NaN to remove all the -999? That would be handy

Indeed it would! Unfortunately I think its literally just an integer that serves as a reminder of what the null value in the datasets is. So we can either use it to set the -9999 (i.e. just refering to ds.Dummy), or we can ignore it and hard-cose the number in.

seastar/oscar/level1.py
ACHMartin: Indeed! Any suggestion how to do it? I would prefer not to have a fixed `.sel(Antenna="Mid")` as we want to keep the modularity for the future. SeaSTAR would have potentially an ATI for the "Mid" antenna

DavidMcCann-NOC: I don't have any immediate ideas really, but if we can think up a way of doing it it'll help out some other functions too.

seastar/oscar/level1.py
DavidMcCann-NOC: Yes - ``CalImage`` is in Mid but not aft or Fore. I can copy the same routine for the fore and aft antennas too? That way we'll get the CalImage in the Fore and Aft datasets but they'll be NaN

ACHMartin:
> > > I did try and solve the -9999 issue, I have a piece of code that works nicely but strangely it gets stuck on a variable called `Dummy` that is actually the identifier for those -9999 values. In this case we have a 1x3 array of -9999 (along the antenna dimension), and when I try and replace the numbers with NaN using the same code that works for all the other arrays it complains about turning a NaN into an int.... I'll have it sorted soon i think
> > 
> > 
> > This sounds a very good surprise. So if I understand correctly, you only need to change this field to NaN to remove all the -999? That would be handy
> 
> Indeed it would! Unfortunately I think its literally just an integer that serves as a reminder of what the null value in the datasets is. So we can either use it to set the -9999 (i.e. just refering to ds.Dummy), or we can ignore it and hard-cose the number in.

What I understand:
ds.Dummy keeps the reference about what a BAD value is, but if we change this value in ds.Dummy it will not change all the BAD values in `ds`.
If the things above is correct, we could do something like:
```
nan_flag = np.unique(ds.Dummy.data)
TODO replace all nan_flag with np.nan
```
so no need to hardly put the -999 in the code.

seastar/oscar/level1.py
ACHMartin: perhaps we could change a bit our logic and filling the missing variables only at the last moment.
So at the step of computing Interferogram and Intensity we put an `if` on the existence of the "Slave" data. If they don't exist, we don't compute these fields.
Then, we send back a new dataset only with the computed fields of interest. All the intermediate fields are not stored in the dataset.
Do you foresee any drawback?

seastar/oscar/level1.py
ACHMartin: Following https://github.com/NOC-EO/seastar_project/pull/224#discussion_r1081019130, I think we should use the fill_missing_variable only as the latest solution. My preference would be to merge only the variables we wish to keep and to merge.

We discuss this in #182, perhaps we should change these function to return a new dataset. Can we merge easily two datasets?

seastar/oscar/level1.py
DavidMcCann-NOC: That will work up until the merge_beams happens and we're back at the same problem of having missing variables for ``.concat``. I think these are the options as they stand (with my understanding of xarray as it is...):

1. Have NaN-filled variables for antennas that can't/shouldnt have those variables computed and keep the Antenna dimension across the beams in one unified dataset
2. Have an AntennaSquint and AntennaATI dimension, requiring changes to functions downstream to get things working again.
3. Have no Antenna dimension in the L1 dataset but store the antenna differences as more variables (i.e. a big long list of variables, CoherenceFore, CoherenceAft, etc.
4. Have no unified L1 dataset but 3 L1 datasets - one for each beam (like we started with right at the beginning). Would require the most changes to the code downstream I think

Personally I think 1 is the easiest but perhaps not the most snazzy, 2 is probably the right way to do things conceptually but will break the processing chain and require some moderate work to get it working again. 3 is 'easy' but very messy and could cause some major headaches and 4 is a no-go i think

seastar/oscar/level1.py
DavidMcCann-NOC: We can easily merge two datasets, but if they're to be along a new dimension they need to contain the same variables. If not along a new dimension then the variables need to be named differently otherwise only variables from one dataset are kept.

It might be a good idea to drop variables we don't want to keep anyway - it will make things easier potentially, but we'll need a list of variables we won't want (or a list that we do want).

DavidMcCann-NOC:
Added the working function to replace the dummy values

seastar/oscar/level1.py
```diff
@@ -759,4 +756,34 @@ def generate_wind_field_from_single_measurement(u10, wind_direction, ds):
...
+        Dataset with variables scrubbed for `dummy_val` and replaced with
+        `replace`
+
+    """
+    for var in ds.data_vars:
+        with warnings.catch_warnings():
+            warnings.simplefilter(action='ignore', category=FutureWarning)
+            bad_val = ds[var].data == dummy_val
+            if np.sum(bad_val) > 0:
+                ds[var] = ds[var].where(~bad_val, replace)
```
ACHMartin, line 782: from https://docs.xarray.dev/en/stable/generated/xarray.where.html I understand `where(cond, x, y)` we get x when cond is True and y when cond is False.
If we have a bad value, we get bad_val=True, then ~bad_value=False, then we do nothing.
Shouldn't we use `where(bad_val, replace)` instead?

seastar/oscar/level1.py
ACHMartin: I would like something between 4. and 1.
Today in the processing, what I copied from you was:
```
1 for track in star_pattern_tracks.keys(): # Loop through star pattern tracks
2    file_index = star_pattern_tracks[track]
3    ds = ss.utils.readers.load_OSCAR_data(oscar_path, file_time_triplets[file_index][1])
4    antenna_ident = ss.utils.tools.identify_antenna_location_from_filename(oscar_path, file_time_triplets[file_index][1])
5    ds = ss.oscar.level1.fill_missing_variables(ds, antenna_ident)
6    print('Processing file index',file_index)
7    # L1 processing
8    for i in list(ds.keys()):
9        ds[i] = ss.oscar.level1.check_antenna_polarization(ds[i])
10        ds[i] = ss.oscar.level1.compute_multilooking_Master_Slave(ds[i], window=7)
11#         ds[i]['Baseline'] = ss.oscar.level1.compute_antenna_baseline(0.2)
12        ds[i] = ss.oscar.level1.compute_antenna_azimuth_direction(ds[i], antenna=antenna_ident[list(ds.keys()).index(i)])
13        ds[i] = ss.oscar.level1.compute_time_lag_Master_Slave(ds[i], options='from_SAR_time')
14        ds[i] = ss.oscar.level1.compute_radial_surface_velocity(ds[i])
15    #Build L1 dataset
16    dsl1[track] = ss.oscar.level1.merge_beams(ds, antenna_ident)
```

I think line 5 can be used later, probably only just before line 16 as we only use one antenna at a time.
Line 10: it will be better if we only get the variable we wants and drop all the Level1-A data (SLC, ...). The output will be a new dataset.
All the line 9, 12-14 can work on this new dataset.
Just before merging, we fill in the missing fields (e.g. interferogram, ...) and we merge.

What do you think about this approach?

seastar/oscar/level1.py
DavidMcCann-NOC: Weirdly, I tried that and it wouldn't work - the ``~`` was a hail-mary that got the results i wanted... So I agree, but I don't know why this works. Doing ``where(cond, x, y)`` kept producing those 'cant do 2D boolean indexing' errors that python seems to love

DavidMcCann-NOC:
> I would like something between 4. and 1.

>Today in the processing, what I copied from you was:

>1 for track in star_pattern_tracks.keys(): # Loop through star pattern tracks
>2    file_index = star_pattern_tracks[track]
>3    ds = ss.utils.readers.load_OSCAR_data(oscar_path, file_time_triplets[file_index][1])
>4    antenna_ident = ss.utils.tools.identify_antenna_location_from_filename(oscar_path, file_time_triplets[file_index][1])
>5    ds = ss.oscar.level1.fill_missing_variables(ds, antenna_ident)
>6    print('Processing file index',file_index)
>7    # L1 processing
>8    for i in list(ds.keys()):
>9        ds[i] = ss.oscar.level1.check_antenna_polarization(ds[i])
>10        ds[i] = ss.oscar.level1.compute_multilooking_Master_Slave(ds[i], window=7)
>11#         ds[i]['Baseline'] = ss.oscar.level1.compute_antenna_baseline(0.2)
>12        ds[i] = ss.oscar.level1.compute_antenna_azimuth_direction(ds[i], antenna=antenna_ident[list(ds.keys()).index(i)])
>13        ds[i] = ss.oscar.level1.compute_time_lag_Master_Slave(ds[i], options='from_SAR_time')
>14        ds[i] = ss.oscar.level1.compute_radial_surface_velocity(ds[i])
>15    #Build L1 dataset
>16    dsl1[track] = ss.oscar.level1.merge_beams(ds, antenna_ident)]

>I think line 5 can be used later, probably only just before line 16 as we only use one antenna at a time.
>Line 10: it will be better if we only get the variable we wants and drop all the Level1-A data (SLC, ...). The output will be a new >dataset.
>All the line 9, 12-14 can work on this new dataset.
>Just before merging, we fill in the missing fields (e.g. interferogram, ...) and we merge.

>What do you think about this approach?

We can chat about this in person - I think I understand and agree - shouldn't require many more changes

seastar/oscar/level1.py
DavidMcCann-NOC: For example - here is with ``where(bad_val, np.NaN)`` :
image
And with ``where(~bad_val, np.NaN)`` :
image

DavidMcCann-NOC:
Added a commit to address what we chatted about on Friday. This works to L1b level now (i think), and is a little more intuitive than it was previously. It also no longer uses up all of my memory which is good.

seastar/oscar/level1.py
```diff
@@ -522,27 +536,32 @@ def compute_antenna_azimuth_direction(ds, antenna):
         head = np.interp(np.reshape(ds.OrbTimeImage, (1, m * n)), ds.GPSTime,
                          ds.OrbitHeading)
         head = np.reshape(head, (m, n))
-        ds['OrbitHeadingImage'] = xr.DataArray.copy(ds.OrbTimeImage, data=head)
-        ds.OrbitHeadingImage.attrs['long_name'] =\
+        OrbitHeadingImage = xr.DataArray.copy(ds.OrbTimeImage, data=head)
```
ACHMartin, line 539: I didn't know about this `.copy`, it looks handy

seastar/oscar/level1.py
```diff
@@ -522,27 +536,32 @@ def compute_antenna_azimuth_direction(ds, antenna):
...
+        OrbitHeadingImage.attrs['long_name'] =\
             'Heading (degrees N) of the airfraft for each pixel in the image'
-        ds.OrbitHeadingImage.attrs['units'] = '[degrees]'
+        OrbitHeadingImage.attrs['units'] = '[degrees]'
```
ACHMartin, line 542: '[degrees]' -> 'degree'

seastar/oscar/level1.py
```diff
@@ -522,27 +536,32 @@ def compute_antenna_azimuth_direction(ds, antenna):
         head = np.interp(np.reshape(ds.OrbTimeImage, (1, m * n)), ds.GPSTime,
                          ds.OrbitHeading)
         head = np.reshape(head, (m, n))
-        ds['OrbitHeadingImage'] = xr.DataArray.copy(ds.OrbTimeImage, data=head)
-        ds.OrbitHeadingImage.attrs['long_name'] =\
+        OrbitHeadingImage = xr.DataArray.copy(ds.OrbTimeImage, data=head)
+        OrbitHeadingImage.attrs['long_name'] =\
```
ACHMartin, line 540: put as 'description'.
as long name, perhaps just Heading from North?

seastar/oscar/level1.py
```diff
@@ -522,27 +536,32 @@ def compute_antenna_azimuth_direction(ds, antenna):
...
             + ds.SquintImage,
             360)
     else:
         raise Exception('SquintImage is a required field in ds')
-    ds.AntennaAzimuthImage.attrs['long_name'] =\
+    AntennaAzimuthImage.attrs['long_name'] =\
         'Antenna azimuth direction for each pixel in the image'
-    ds.AntennaAzimuthImage.attrs['units'] = '[degrees North]'
-    return ds
+    AntennaAzimuthImage.attrs['units'] = '[degrees North]'
```
ACHMartin, line 560: without [ ] and 's'

seastar/oscar/level1.py
```diff
@@ -522,27 +536,32 @@ def compute_antenna_azimuth_direction(ds, antenna):
...
-            ds.OrbitHeadingImage
+        AntennaAzimuthImage = np.mod(
+            OrbitHeadingImage
             + look_direc_angle[lookdirec]
             + ds.SquintImage,
             360)
     else:
         raise Exception('SquintImage is a required field in ds')
-    ds.AntennaAzimuthImage.attrs['long_name'] =\
+    AntennaAzimuthImage.attrs['long_name'] =\
```
ACHMartin, line 558: this is the "description"; long_name = "Antenna azimuth direction"

seastar/oscar/level1.py
```diff
@@ -551,59 +570,87 @@ def compute_time_lag_Master_Slave(ds, options):
...
+            TimeLag.attrs['description'] =\
+                'Time difference between antenna master and slave.'
+            TimeLag.attrs['units'] = 's'
+        else:
+            TimeLag = xr.DataArray(data=np.NaN)
+            TimeLag.attrs['long_name'] = 'Time lag'
+            TimeLag.attrs['description'] =\
+                'Time difference between antenna master and slave.'\
+                ' Set to NaN as no slave data in beam dataset.'
+            TimeLag.attrs['units'] = 's'
```
ACHMartin, line 601: is there any meaning to calculate a timelag if there is no Slave?
Should we raise an error instead?

If it is from_aircraft_velocity, we don't have this issue and what you propose looks good to me.

seastar/oscar/level1.py
```diff
@@ -551,59 +570,87 @@ def compute_time_lag_Master_Slave(ds, options):
...
+        TimeLag = (ds.Baseline / 2 * ds.MeanForwardVelocity)
+        TimeLag.attrs['long_name'] = 'Time lag'
+        TimeLag.attrs['description'] =\
+            'Time difference between antenna master and slave.'
+        TimeLag.attrs['units'] = 's'
+    return TimeLag
 
 
-def compute_radial_surface_velocity(ds):
+def compute_radial_surface_velocity(ds_L1a, ds_ml):
```
ACHMartin, line 611: I would be in favour to have only one DataSet as input. To be discussed.

ACHMartin (commented):
Thanks. It is a good way to revisit all our coding from the early start. We have already done a nice journey...

seastar/oscar/level1.py
DavidMcCann-NOC: There are only two variables used from ds_L1a (wavenumber and antenna azimuth), however these aren't in the ds_ml dataset. So either these can get passed into the ds_ml dict of beam datasets beforehand or ds_L1b can start earlier up the chain as a dict of datasets (i.e., replacing ds_ml) and then get converted into a merged-beam dataset at the end of the L1 processing chain.

I've kept them separate so far so that the L1b dataset doesn't start life as one thing and then turn into another, but if you don't think it matters too much then that would be an easy and clean fix, so only ds_L1b would get passed into the RSV computation. Otherwise the other options are:
- variables copied into the multilooking dataset
- variables passed explicitly in the function header (probably cleaner?)

ACHMartin: It is quite normal to have repetition of variables between level products if needed in the processing chain.

L1A -> processing -> outputs in L1B
L1A -> push useful variables to -> L1B

.wawenumber can probably be an attribute instead of a variable as it is constant. I am not sure it will make our life easier. I fear the opposite. Perhaps we can keep it without dimension.

DavidMcCann-NOC: Sorted in the last commit (i think). I just passed the needed variables over the line before calling the function and now ds_ml is the only dataset passed into the function input

seastar/oscar/level1.py
ACHMartin: Before merging, could we review how the full processing chain has evolved from what we had originally (see below):
Thanks
> ```
> 1 for track in star_pattern_tracks.keys(): # Loop through star pattern tracks
> 2    file_index = star_pattern_tracks[track]
> 3    ds = ss.utils.readers.load_OSCAR_data(oscar_path, file_time_triplets[file_index][1])
> 4    antenna_ident = ss.utils.tools.identify_antenna_location_from_filename(oscar_path, file_time_triplets[file_index][1])
> 5    ds = ss.oscar.level1.fill_missing_variables(ds, antenna_ident)
> 6    print('Processing file index',file_index)
> 7    # L1 processing
> 8    for i in list(ds.keys()):
> 9        ds[i] = ss.oscar.level1.check_antenna_polarization(ds[i])
> 10        ds[i] = ss.oscar.level1.compute_multilooking_Master_Slave(ds[i], window=7)
> 11#         ds[i]['Baseline'] = ss.oscar.level1.compute_antenna_baseline(0.2)
> 12        ds[i] = ss.oscar.level1.compute_antenna_azimuth_direction(ds[i], antenna=antenna_ident[list(ds.keys()).index(i)])
> 13        ds[i] = ss.oscar.level1.compute_time_lag_Master_Slave(ds[i], options='from_SAR_time')
> 14        ds[i] = ss.oscar.level1.compute_radial_surface_velocity(ds[i])
> 15    #Build L1 dataset
> 16    dsl1[track] = ss.oscar.level1.merge_beams(ds, antenna_ident)
> ```

PR #228, opened 2023-02-01
Mask generation problem

DavidMcCann-NOC:
This works on my machine after the L1 processing chain, invoke with something like:

``mask, coast_polygons = ss.utils.tools.compute_land_mask_from_GSHHS(ds_L1b['Track_1'])``

Timed at 157s using data from the 17th may

Added some print commands to inform of how far along it is.


ACHMartin:
No comment. I hope it will work with mine as well. Thanks

PR #233, opened 2023-02-03
New, faster land mask computation

DavidMcCann-NOC:
Takes a DataArray as input - must have `latitude` and `longitude` coords

Should work out of the box - use it like:
`mask = ss.utils.tools.compute_land_mask_from_GSHHS(ds_L1b[track].Interferogram.sel(Antenna='Mid'), skip=1/1000)`

Note that it might not like the Antenna dimension so it has to be selected out on input

### DIFF
--- a/seastar/utils/readers.py
+++ b/seastar/utils/readers.py
@@ -40,6 +40,44 @@ def _set_file_paths():
     return local_paths
 
 
+def _read_DAR_config(date):
+    """
+    Read DAR track name config.
+
+    Reads the seastarex_DAR_config.ini file containing names for aquisition
+    as a dict with keys equalling track names and values equalling the file
+    list number for loading.
+
+    Parameters
+    ----------
+    date : ``str``, ``int``
+        Date of aquisition in the form YYYYMMDD
+
+    Raises
+    ------
+    Exception
+        DESCRIPTION.
+
+    Returns
+    -------
+    DAR_tracks : ``dict``
+        Dict of {track names : file number}. File numbers of type ``int``.
+        Track names capitalised.
+
+    """
+    if type(date) is not str:
+        date = str(date)
+    config = ConfigParser()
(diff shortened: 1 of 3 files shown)


seastar/utils/readers.py
```diff
@@ -40,6 +40,44 @@ def _set_file_paths():
     return local_paths
 
 
+def _read_DAR_config(date):
```
ACHMartin, line 43: I don't really see the value of the config file as it is. Perhaps it could be a good way to get the triplet values.
but there is no harm to keep it

seastar/utils/tools.py
```diff
@@ -602,16 +603,28 @@ def compute_land_mask_from_GSHHS(ds, erosion=False, boundary=None,
             count = count + 1
             if not int(np.mod(count, ((m*n) / 10))):
                 print(int((count / (m*n)) * 100), '% complete')
-
             for k in coast_polygons.keys():
                 mask[i, j] = mask[i, j] or\
-                    Point(ds.longitude.data[i, j], ds.latitude.data[i, j])\
-                    .within(coast_polygons[k])
+                   Point(longitude.data[i, j], latitude.data[i, j])\
```
ACHMartin, line 608: why longitude.data? I understand longitude is a ndarray

ACHMartin (approved):
please change the longitude/latitude to lon_skip or something similar to make it easier to read

seastar/utils/readers.py
DavidMcCann-NOC: Its convenient for me when switching between days for processing, although admittedly there aren't many days to switch between right now.

seastar/utils/tools.py
DavidMcCann-NOC: Fair point - I assumed i'd left the .data in because Point wouldn't work with the DataArray output of indexing a coordinate directly but it turns out it does just fine

PR #238, opened 2023-02-09
Merge elements for cost minimisation link to SciReC performance assessment + unit test

ACHMartin:
Addition and modification of a wide range of functions to minimize cost function in retrieval/cost_function + the dependencies in other package.
First attempt of a structure for the processing chain for SciReC performance assessment.
Implementation of first unit tests


seastar/performance/scene_generation.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+
+    truth_out['Sigma0'] = ss.gmfs.nrcs.compute_nrcs(inst, geo, gmf.nrcs)
+
+    # TODO below, to enable to compute this without a loop through the antenna;
+    for aa, ant in enumerate(level1.Antenna.data):
+        model_rvl_list[aa] = ss.gmfs.doppler.compute_total_surface_motion(inst.sel(Antenna=ant), geo,
+                                                                          gmf=gmf.doppler.name)
+    truth_out['RVL'] = xr.concat(model_rvl_list, dim='Antenna')
+
```
DavidMcCann-NOC, line 99: To be done

seastar/performance/scene_generation.py
```diff
@@ -0,0 +1,333 @@
...
+import seastar
+from seastar.utils.tools import dotdict
+
+# import seastar.gmfs.doppler
+import pdb
+
+
+def create_scene_dataset(geo, inst):
+    """
+    Scene generation function of geophysical parameters (direct models) and acquisition geometry
```
DavidMcCann-NOC, line 20: Would we want the output to be potentially automatically shaped by the input data? I.e. if we have  instrument characteristics as a constant or a constant along-swath but geo data for every pixel then the function could shape the instrument data automatically to suit (build an array from a constant, or a vector along-swath), or alternatively if the geo data is just a single value but the instrument data is more than this then the function could pick the lowest-common-denominator and use instrument characteristics from that single pixel.

Might be a little complicated, but it would be great for the default behaviour of the function to 'just run', with options to interpolate or otherwise build arrays from single point data if wanted. 

Perhaps the options could be related to how you want the output, i.e. 'output_like=inst'

In practice this will likely need a bunch of conditional logic i suppose

seastar/performance/scene_generation.py
```diff
@@ -0,0 +1,333 @@
...
+        uncertainty on: NRCS, (Doppler? or RVL?)
+
+
+    Returns
+    -------
+    level1 : ``xarray.Dataset``
+        L1 measurements dataset with data: Sigma0 (NRCS), RVL (Radial Velocity) <= truth + noise
+        and coordinates: longitude, latitude, antenna, incidence angle, azimuth look direction
+        and attributes: (if needed)
+        L1.shape (antenna, across, along)
```
DavidMcCann-NOC, line 48: From above, the trickiest part to being adaptive about the output will likely come from the dims and coords we use for the output dataset. Simplified if the inputs are DataArrays with them already in but it would be good for the function to happily make them itself from lat long antenna etc data that came in the inputs.

Sometimes dims and coords arent set properly and most of the dataset combination problems we've been having in the L2 processor stem from that. A solution we implement here could be very useful elsewhere

seastar/performance/scene_generation.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+
+def uncertainty_fct( truth, uncertainty):
+    """
+    Computed and Filled in all the uncertainty parameters
+
+    Should work for:
+    - a point
+    - same dimension for truth and uncertainty 1D, 2D, xD (with antenna?)
```
DavidMcCann-NOC, line 115: This is an interesting problem i think, and linked to how we implement the other functions here. Probably worth a look at how other people usually deal with datasets with same/different dims etc.

DavidMcCann-NOC (approved):
Big merge this one! And lots of functions that I won't be using soon. So happy to merge and pick up any small bugs if they arise.

seastar/performance/scene_generation.py
ACHMartin: The difficulty is to related the two elements together.
Do we expect both geo and inst are on a grid of 1km resolution?
But if we use real wind/current model data for geo, the grid will certainly not be on a 1km grid. Do we resample on 1km grid?

seastar/performance/scene_generation.py
ACHMartin: I am not sure to understand. I assume the inputs will be Dataset with the correct format. Potentially, geo and inst might not be aligned.

Issue #240, opened 2023-02-10
Standardise variable names for velocity components


DavidMcCann-NOC:
Standardize the variable names for velocity components (currents and wind) across all functions so that we have:
Currents:
`CurrentU`
`CurrentV`
Wind:
`WindU`
`WindV`

Also check and update all docstrings

ACHMartin:
Also change CurrentMagnitude to CurrentVelocity

ACHMartin:
we also have vis_u, vis_v, vis_speed, vis_direction which correspond to the Visible Wind (relative to the surface), on contrast to the Wind which is an absolute wind compared to solid earth.
Proposition to VisU, VisV, VisSpeed, VisCurrent (not ideal...), VisibleU, ...
RelativeWindU, ...
RelWindU, ...

Do we move our previous WindU to AbsWindU, and our visible to RelWindU, ...

ACHMartin:
This raise, lots of question as for exemple compute_nrcs is using WindSpeed and WindDirection as input.
Same for compute_wasv I guess.

Then should we move all these functions to RelativeWindSpeed, ...

DavidMcCann-NOC:
In my view 'absolute' has a defined mathematical meaning, i.e. `sqrt( Windspeed **2 )`, so as long as AbsoluteWindSpeed satisfies this then yes, certainly. 'relative' denotes being relative to something else, and as its attached to 'speed' then RelativeWindSpeed would denote the wind speed relative to a different speed or reference frame. In aeronautics they use 'true' and 'indicated' to basically mean 'absolute' and 'relative', with the speed its relative to being the sensor (aircraft).

If the measurement is from, or simulates, a fixed sensor or reference frame (i.e., Eulerian) then in my view its `Absolute`. If its from a moving sensor or reference frame and is un-corrected (i.e., true Lagrangian) and/or the product or difference of another velocity vector then its `Relative`

DavidMcCann-NOC:
> we also have vis_u, vis_v, vis_speed, vis_direction which correspond to the Visible Wind (relative to the surface), on contrast to the Wind which is an absolute wind compared to solid earth.

I really should look further up shouldn't I! I'm not sure I'm up for visible wind, but maybe thats because i've never used that terminology before and to me it sound like i can see the wind!

But if its standard in scatterometry, meterology etc then would be happy with that definition. 

Perhaps finding a standard that other people use and following that might help? Personally I'd probably get along with any variable change as long as its documented

[...]

PR #242, opened 2023-02-15
Few different changes

DavidMcCann-NOC:


DavidMcCann-NOC:
I image a few of these changes will be superseded by recent developments in #240

seastar/utils/readers.py
```diff
@@ -127,7 +128,7 @@ def readNetCDFFile(netCFD_path):
     :type netCFD_path: ``str``
 
     :raises: ``ValueError`` if file cannot be read as netCDF and \
-    returns ``None`` object
+        returns ``None`` object
```
ACHMartin, line 131: In this file there is the "old" documentation style using :param for many function, as for example the FindNetCDFilepaths...
In addition, are these functions specific to OSCAR or could they be used for some other things?
I believe findNetCDF is quite universal, but findTripplet is not. Perhaps it could be moved to the Oscar package with a new files readers/tools/...

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -63,6 +73,21 @@
...
+        ncwriteatt([file_path, file_name], 'SquintImage','description','Beam squint along the ground for each pixel in the image')
+        
+        SquintMounted=rad2deg(SquintMounted);
+        [m,n] = size(SquintMounted);
+        nccreate([file_path, file_name],'SquintMounted','Dimensions',...
+            {info.Dimensions(find(cell2mat(dim_list(:,1)) == m)).Name, m,...
+            info.Dimensions(find(cell2mat(dim_list(:,1)) == n)).Name, n},...
+            'Format','netcdf4')
+        ncwrite([file_path, file_name], 'SquintMounted', SquintImage)
+        ncwriteatt([file_path, file_name], 'SquintMounted','long_name','LOS squint')
```
ACHMartin, line 87: For the long name, I think it will be better that it represents the same name as the variable. In this case "Mounted squint".
An other solution might be to change  the variable "SquintMounted" to "SquintLOS", but I don't see any value in this

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -63,6 +73,21 @@
...
+        ncwriteatt([file_path, file_name], 'SquintMounted','units','deg')
+        ncwriteatt([file_path, file_name], 'SquintMounted','description','Beam squint along the line of sight for each pixel in the image')
```
ACHMartin, line 89: Change to 'Beam squint along the line of sight for each pixel in the image. Also called Mounted squint angle.'

Ideally we would like to add a schematic somewhere for this angle. Do you have any idea where we could put it? Then we could refer to this schematic.

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -76,12 +101,19 @@
             info.Dimensions(find(cell2mat(dim_list(:,1)) == n)).Name, n},...
             'Format','netcdf4')
         ncwrite([file_path, file_name], 'OrbitHeadingImage', OrbitHeadingImage)
+        ncwriteatt([file_path, file_name], 'OrbitHeadingImage','long_name','Orbit heading')
+        ncwriteatt([file_path, file_name], 'OrbitHeadingImage','units','deg')
+        ncwriteatt([file_path, file_name], 'OrbitHeadingImage','description','Heading relative to North of the sensor for each pixel in the image')
```
ACHMartin, line 106: North of the sensor? Is it where the sensor is pointing, or is it linked to the aircraft heading?

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -76,12 +101,19 @@
...
+
         [m,n] = size(OrbitYawImage);
         nccreate([file_path, file_name],'OrbitYawImage','Dimensions',...
             {info.Dimensions(find(cell2mat(dim_list(:,1)) == m)).Name, m,...
             info.Dimensions(find(cell2mat(dim_list(:,1)) == n)).Name, n},...
             'Format','netcdf4')
         ncwrite([file_path, file_name], 'OrbitYawImage', OrbitYawImage)
+        ncwriteatt([file_path, file_name], 'OrbitYawImage','long_name','Orbit yaw')
+        ncwriteatt([file_path, file_name], 'OrbitYawImage','units','deg')
+        ncwriteatt([file_path, file_name], 'OrbitYawImage','description','Yaw of sensor relative to platform heading along the ground for each pixel in the image')
```
ACHMartin, line 116: Is it sensor, gimbal or platform yaw?

matlab/metasensing/squ_from_netcdf.m
```diff
@@ -24,8 +26,23 @@
 
 squintx=round(squintx.*pi/180*10000.)./10000;
 
+if SystemType=='M'
+ squintg=round(squintg.*pi/180*10000.)./10000;
```
ACHMartin, line 30: in python, I know you can define the rounding in term of digit ie `round(square, 4)` for 4 digits.
I guess in Matlab something similar should exist

ACHMartin (approved):
great, small comments, ready to go.
You might open a new issue for mowing files, so you can close the PR rapidly.
Thanks

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: Thats not a bad idea. If these scripts are going into the documentation then I assume we can insert a .png image pretty simply

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: Good point. Probably better as 'Sensor heading relative to North'. Its important that North is in there as its a heading

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: I think platform actually. But I'll have to dig in to the code / old emails to check

matlab/metasensing/squ_from_netcdf.m
DavidMcCann-NOC: Fair, although this is MetaSensing's coding here

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
ACHMartin: I guess so, but where to show it. Could we have an overview/introduction for the OSCAR package? It would be the ideal place to put it I guess.

DavidMcCann-NOC: Of course - I think thats a good idea. The docs can be anything we want format-wise really, the auto-generated documentation is only called for specific places, the rest is typeset

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
ACHMartin: it's not the same for all "Orbit". For the comment before, you wrote it was for the sensor not the platform.
I think all these attitude measurements are normally referring to the gimbal (it is where the IMU is mounted).

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
ACHMartin: which file should I touch? In docs/source I see some package files with .rst, but not all.
Where are the generated "html"? How are they transferred to "https://seastar-project.readthedocs.io/en/latest/"

seastar/utils/readers.py
DavidMcCann-NOC: Ah yes, lots of old functions in there. Sphinx doesn't mind about the difference in docstring format (its still valid), but i'll get them changed as part of my docstring updating issue so that everything is in the same style for readability within the modules.

Some functions are specific to OSCAR, so i agree a better home for them would be in the oscar sub-package

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: This is part of my commits I think - all of the recent sphinx stuff I've been working on will be merged.

I'm also trying to figure out how the static builds get linked and hosted by github, that way we can have the html linked through properly on the repository page

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: Should I change this to 'yaw of gimbal relative to platform heading'?

ACHMartin: Something like: "Remaining uncorrected yaw of the antennas system on the gimbal. The yaw is relative to heading"
Does it make sense like this?

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
ACHMartin: OrbitHeading: Is it the sensor (antennas beam), the gimbal, or the platform?
Do we have different values per Antenna?

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
ACHMartin: Do we have different values for each Antennas, or is it just reprocessed on a different grid?

DavidMcCann-NOC: > Do we have different values for each Antennas, or is it just reprocessed on a different grid?
I'll have to check that - sorry, I missed these comments before I merged it yesterday - I'll add a note to my attribute writing issue to make sure these are updated as well.


> Something like: "Remaining uncorrected yaw of the antennas system on the gimbal. The yaw is relative to heading"
> Does it make sense like this?

Yeah that makes sense. I'll add this in, thanks

Issue #180, opened 2022-10-21
remove specific mention of CrossRange, GroundRange


ACHMartin:
Systematically remove the specific mention of CrossRange, GroundRange

For example in gmfs/doppler/compute_wasv:
ds_wa = xr.DataArray(wasv_rsv.data, coords=(L1.CrossRange, L1.GroundRange))

or Oscar/level1/generate_wind_field_from_single_measurement

DavidMcCann-NOC:
Just going through this now, sorted the mention in compute_wasv. The other instances relate to generating meshed X, Y grids - I'll find a better way of doing it to scrub the mentions of these Coords

DavidMcCann-NOC:
Closed with #245

PR #245, opened 2023-02-17
Issue #180

DavidMcCann-NOC:
Removed explicit mentions of CrossRange and GroundRange from all remaining functions, replaced with pointers to dims of relevant DataArrays in functions

Also removed generate_wind_field_from_single_measurement() from retrieval\level2


seastar/oscar/level1.py
```diff
@@ -74,14 +74,20 @@ def merge_beams(ds_dict, antenna_id):
     ds_level1 = ds_level1.assign_coords(Antenna=('Antenna', antenna_id))
     key_list = list(ds_dict.keys())
     ds_level1.coords['latitude'] = xr.merge(
-            [ds_dict[key_list[0]].LatImage.dropna(dim='CrossRange'),
-             ds_dict[key_list[1]].LatImage.dropna(dim='CrossRange'),
-             ds_dict[key_list[2]].LatImage.dropna(dim='CrossRange')],
+            [ds_dict[key_list[0]].LatImage
+             .dropna(dim=ds_dict[key_list[0]].LatImage.dims[0]),
```
ACHMartin, line 78: How to be sure dims[0] will be 'CrossRange'? Should we drop it for all dimension (even if it is not needed)? Or another solution is to define the dimension in the header of the function.

seastar/oscar/level1.py
```diff
@@ -237,12 +243,13 @@ def compute_multilooking_Master_Slave(ds, window=3, vars_to_send=['Intensity', '
         ds = compute_SLC_Master_Slave(ds)
     if 'SigmaSLCSlave' not in ds.data_vars:
         ds_out['IntensityAvgComplexMasterSlave'] = (ds.SigmaSLCMaster ** 2)\
-            .rolling(GroundRange=window).mean()\
-            .rolling(CrossRange=window).mean()
+            .rolling({ds.SigmaSLCMaster.dims[1]: window}).mean()\
```
ACHMartin, line 246: We should add in the description of the function, ds should have the minimum following fields (SigmaReal, ...) + the optional fields (SigmaReal..Slave) and they need to be 2D fields.
Probably good to check it at the beginning and raise an issue if not the case.

Perhaps with opening a new issue

seastar/oscar/level1.py
```diff
@@ -252,16 +259,14 @@ def compute_multilooking_Master_Slave(ds, window=3, vars_to_send=['Intensity', '
...
-        #     np.full(ds_out.IntensityAvgComplexMasterSlave.shape, np.NaN))
+
         ds_out['Interferogram'] = xr.DataArray(data=np.NaN)
         ds_out.Interferogram.attrs['description'] =\
             'Interferogram between master/slave antenna pair.'\
             ' Values set to NaN as no Slave data present in beam dataset'
     else:
         ds_out['Interferogram'] = (
-            ['CrossRange', 'GroundRange'],
+            [ds.SigmaSLCMaster.dims[0], ds.SigmaSLCMaster.dims[1]],
```
ACHMartin, line 269: `[ds.SigmaSLCMaster.dims]` should be sufficient, no need for [0] and [1]. If we check the size at the beginning

ACHMartin (approved):
good to go, you might wish to trigger new issues to the point I raised on dims for the docs and to test its 2d fields

seastar/oscar/level1.py
DavidMcCann-NOC: You're probably not wrong here - might be best to dropna for all dimensions just in case, won't hurt I don't think. Plus we've already had the dims switch order for an unknown reason so best to get ahead of a bug

seastar/oscar/level1.py
DavidMcCann-NOC: Ahh yes, good catch. I had just replaced what was there previously but that would be a better way to proceed

PR #250, opened 2023-03-16
small modifications

ACHMartin:


seastar/utils/tools.py
```diff
@@ -120,6 +120,41 @@ def windUV2SpeedDir(u, v):
     wdir = np.mod(-90 - np.angle(tmp) * 180 / np.pi, 360)
     return wspd, wdir
 
+
+def EarthRelativeSpeedDir2all(ds):
```
DavidMcCann-NOC, line 124: Potentially opaque function name ('...2all'). Not entirely sure what it should be changed to though...

Useful function though

ACHMartin: I fully agree!

DavidMcCann-NOC (approved):
No real comments, it can go as it I think. It sounds like the add_central_electromagnetic_wavenumber function would be better to return a DataArray in this instance - probably worth adding an issue to update this? Its an old function from when we were passing around datasets more frequently.

PR #258, opened 2023-03-30
Add IntensityAvgMaster field output for Oscar/level1 even when there is no Slave

ACHMartin:
in "oscar/level1/compute_multilooking_Master_Slave", modification on IntensityAvgMaster to have this field compute when there is a Slave field (case before) or not.

In the typical SeaSTAR/Ascat geometry with ATI on squinted beams and standard SAR on mid beam, the IntensityAvgMaster will now be available for fore/aft and mid. For the mid beam, Intensity==IntensityAvgMaster

### DIFF
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -309,6 +309,18 @@ def compute_multilooking_Master_Slave(ds, window=3,
             'Interferogram between master/slave antenna pair.'
     ds_out.Interferogram.attrs['long_name'] = 'Interferogram'
     ds_out.Interferogram.attrs['units'] = 'rad'
+
+    ds_out['IntensityAvgMaster'] = np.abs(
+        (ds.SigmaSLCMaster ** 2) \
+            .rolling({ds.SigmaSLCMaster.dims[1]: window}).mean() \
+            .rolling({ds.SigmaSLCMaster.dims[0]: window}).mean()
+    )
+    ds_out.IntensityAvgMaster.attrs['long_name'] = \
+        'Intensity Master'
+    ds_out.Intensity.attrs['description'] = \
+        'Average absolute single look complex image intensity for Master SLC (|M^2|)'
+    ds_out.IntensityAvgMaster.attrs['units'] = ''
+
     if 'SigmaSLCSlave' in ds.data_vars:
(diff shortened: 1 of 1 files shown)


seastar/oscar/level1.py
```diff
@@ -311,31 +310,38 @@ def compute_multilooking_Master_Slave(ds, window=3,
...
+            (ds.SigmaSLCSlave ** 2)\
+                .rolling({ds.SigmaSLCSlave.dims[1]: window}).mean()\
+                .rolling({ds.SigmaSLCSlave.dims[0]: window}).mean()
+        )
         ds_out['Coherence'] =\
             ds_out.Intensity / np.sqrt(ds_out.IntensityAvgMaster
                                        * ds_out.IntensityAvgSlave)
     else:
-        ds_out['IntensityAvgMaster'] = xr.DataArray(data=np.NaN)
+        # ds_out['IntensityAvgMaster'] = xr.DataArray(data=np.NaN)
```
DavidMcCann-NOC, line 322: i -think- this will be ok as we do that `fill_missing_variables` trick before merging the beams but I'll have to test it. xarray gets grumpy when datasets don't contain the same variables when concatenating. It may be that we had to set it to NaN to just make that part of the workflow work.

DavidMcCann-NOC (commented):
This should all work fine - I'll give it a quick test before merging. Thanks!

seastar/oscar/level1.py
```diff
@@ -311,31 +310,38 @@ def compute_multilooking_Master_Slave(ds, window=3,
...
+    ds_out.IntensityAvgSlave.attrs['long_name'] = \
+        'Intensity Slave'
+    ds_out.IntensityAvgSlave.attrs['units'] = ''
     ds_out.Coherence.attrs['long_name'] = \
         'Coherence'
     ds_out.Coherence.attrs['description'] = \
         'Coherence between master/slave antenna pair'
     ds_out.Coherence.attrs['units'] = ''
+
+    ds_out['IntensityAvgMaster'] = np.abs(
```
DavidMcCann-NOC, line 335: This block needs to move to line 317 - i.e., before the `Coherence` is computed, as it relies on `ds_out.IntensityAvgMaster`

Issue #264, opened 2023-06-07
Combined retrieval crashing due to x0 being out of bounds


DavidMcCann-NOC:
The simultaneous retrieval works for some cases but not all, crashing due to the initial guess `x0` being out of bounds. For some cases (e.g., Track_2 from 20220517) it is due to the windspeed being > 30m/s, for some (e.g., Track_15 from 20220522) it is due to a current vector component being >5m/s.

There are either issues with the implementation of `x0`, `bounds`, or the input data. Numerous attempts at cleaning up the input data have been attempted but there are still crashes. Cleanup attempts have included:

- Removing 'edge effects' from data variables by eroding the data by 2 pixels
- Removing land using `ss.utils.tools.compute_land_mask_from_GSHHS`

ACHMartin:
use try and catch to solve the bug in order not to crash the full processing, but just the pixel.
@DavidMcCann-NOC could you please send me the full error

DavidMcCann-NOC:
Closed with #267

ACHMartin:
David, I will keep the issue open. I need to work on it for this x0 out of bound to never happen in the first place. It is an issue in my initialisation. It is on me.

Issue #259, opened 2023-03-30
Pytest failing for test_level2.py


DavidMcCann-NOC:
Pytest currently does not run for `test_level2` because of the formatting of the input data to the test functions. Additionally, there needs to be a check before the test functions to remove/ignore any dimensions, indexes, coordinates and data vars in the l2 output of `wind_current_retrieval` so that the input test dataset and the l2 output have the same form for comparison

ACHMartin:
the test_level2 failed because of multiprocessing
seastar\tests\retrieval\test_level2.py:166:
# Test on full xr.DataSet with 2 dimension
    l2, lmout = level2.wind_current_retrieval(L1, N, gmf_mouche, ambiguity)

PR #267, opened 2023-06-13
Issue #264

DavidMcCann-NOC:
Changes to address Issue #264 causing crashes of find_minima in the case of out-of-bounds x0. 

In test cases with un-physical RSV and Sigma0 this fix does not always flag bad boundary conditions, however it does prevent the function from crashing during parallel computation.

### DIFF
--- a/seastar/retrieval/ambiguity_removal.py
+++ b/seastar/retrieval/ambiguity_removal.py
@@ -77,7 +77,8 @@ def ambiguity_sort_by_cost(lmout):
     index : ``xarray.DataArray``
         a dataArray index. Can be applied directly to lmout with lmout.isel(Ambiguities=index_cost)
     """
-    index = lmout.cost.argmin(dim='Ambiguities')
+    cost = xr.where(np.isnan(lmout.cost), np.inf, lmout.cost)
+    index = cost.argmin(dim='Ambiguities')
     return index
 
 
--- a/seastar/retrieval/cost_function.py
+++ b/seastar/retrieval/cost_function.py
@@ -10,6 +10,7 @@ import xarray as xr
 from scipy.optimize import least_squares
 import seastar
 from seastar.utils.tools import dotdict, da2py
+import sys
 
 # import seastar.gmfs.doppler
 # import pdb # pdb.set_trace() # where we want to start to debug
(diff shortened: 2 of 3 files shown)


seastar/retrieval/cost_function.py
```diff
@@ -142,6 +149,40 @@ def find_minima(level1_pixel, noise_pixel, gmf):
...
+        Dict output from scipy.optimize.least_squares.
+
+    Returns
+    -------
+    lmout_null : ``dict``
+        copy of `lmout_dict` with variables changed to NaN.
+
+    """
+    lmout_null = lmout_dict.copy()
+    for var in lmout_dict.keys():
```
ACHMartin, line 174: very nice, very compact. You're a coding in a nice pythonic way now :)

seastar/retrieval/cost_function.py
```diff
@@ -142,6 +149,40 @@ def find_minima(level1_pixel, noise_pixel, gmf):
...
+        if isinstance(lmout_dict[var], np.ndarray):
+            lmout_null[var] = np.full(lmout_dict[var].shape, np.NaN)
+        else:
+            lmout_null[var] = np.NaN
+
+    lmout_null['message'] = 'x0 found to be infeasible and out of bounds'
```
ACHMartin, line 180: perfect

ACHMartin (approved):
I think it will be a good habit to change systematicaly the _version.py. I know it is a bit useless now, as you didn't run anything, So up to you.
Thanks, nice work.

seastar/retrieval/cost_function.py
```diff
@@ -10,6 +10,7 @@
 from scipy.optimize import least_squares
 import seastar
 from seastar.utils.tools import dotdict, da2py
+import sys
```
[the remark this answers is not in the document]
DavidMcCann-NOC: Nope, must have been left in from a test. Deleted now, thanks

seastar/retrieval/cost_function.py
DavidMcCann-NOC: Come a long way from Matlab!

seastar/retrieval/ambiguity_removal.py
```diff
@@ -77,7 +77,8 @@ def ambiguity_sort_by_cost(lmout):
     index : ``xarray.DataArray``
         a dataArray index. Can be applied directly to lmout with lmout.isel(Ambiguities=index_cost)
     """
-    index = lmout.cost.argmin(dim='Ambiguities')
```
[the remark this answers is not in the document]
DavidMcCann-NOC: Ah, I didn't realise that was an available option. Probably better to use that, thanks

DavidMcCann-NOC:
> > .
> 
> I assume you did some test. Any chance you could push them as unit_test? If too complicated to do it rapidly (half a day), I suggest you put your code for the test in a new issue and we will clean that up by then

I'm not sure I understand - do you mean push my testing notebook for this issue but on the unit_test branch?

ACHMartin:
> > > .
> > 
> > 
> > I assume you did some test. Any chance you could push them as unit_test? If too complicated to do it rapidly (half a day), I suggest you put your code for the test in a new issue and we will clean that up by then
> 
> I'm not sure I understand - do you mean push my testing notebook for this issue but on the unit_test branch?

We have unit test in "tests/...". It will be good to add the tests you have done (quick test in a matter of seconds or less than 1 minute) in the unit_test, so we can test it systematically.
If you don't have time to do it now, raise an issue with the link to your notebook

DavidMcCann-NOC:
> > > > .
> > > 
> > > 
> > > I assume you did some test. Any chance you could push them as unit_test? If too complicated to do it rapidly (half a day), I suggest you put your code for the test in a new issue and we will clean that up by then
> > 
> > 
> > I'm not sure I understand - do you mean push my testing notebook for this issue but on the unit_test branch?
> 
> We have unit test in "tests/...". It will be good to add the tests you have done (quick test in a matter of seconds or less than 1 minute) in the unit_test, so we can test it systematically. If you don't have time to do it now, raise an issue with the link to your notebook

This might take more time than that as there is an open unit_test issue (Issue #259) that potentially needs to be fixed first

ACHMartin:
> > > > > .
> > > > 
> > > > 
> > > > I assume you did some test. Any chance you could push them as unit_test? If too complicated to do it rapidly (half a day), I suggest you put your code for the test in a new issue and we will clean that up by then
> > > 
> > > 
> > > I'm not sure I understand - do you mean push my testing notebook for this issue but on the unit_test branch?
> > 
> > 
> > We have unit test in "tests/...". It will be good to add the tests you have done (quick test in a matter of seconds or less than 1 minute) in the unit_test, so we can test it systematically. If you don't have time to do it now, raise an issue with the link to your notebook
> 
> This might take more time than that as there is an open unit_test issue (Issue #259) that potentially needs to be fixed first

I think this can be done despite of this. Pytest can fail on different items (unit test), it doesn't prevent to add a new unit test and running it (I think).
Issue #259 is not very clear to me.  Anyway, I think you don't have much time, but it will be a pity to lose the testing you have done. Keeping it in a new issue with the associated notebooks will certainly help to do the unit test for the try/catch in the future.

PR #272, opened 2023-07-27
Fix coherence

elemerle:
Hi guys, 
Now that I am a pro in PyCharm, Git GitHub and everything, I am doing my fist pull request ;)

In this branch I corrected the Coherence calculation by correcting the parenthesis thing in IntensityAvgMaster and IntensityAvgSlave variables. I also corrected/added metadata for those two variables.

Cheers,

Eva

### DIFF
--- a/_version.py
+++ b/_version.py
@@ -6,5 +6,5 @@ Version of the codebase
 
 """
 
-__version__ = '2023.06.2'
+__version__ = '2023.07.1'
 
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -310,11 +310,11 @@ def compute_multilooking_Master_Slave(ds, window=3,
     ds_out.Interferogram.attrs['long_name'] = 'Interferogram'
     ds_out.Interferogram.attrs['units'] = 'rad'
 
-    ds_out['IntensityAvgMaster'] = np.abs(
-        (ds.SigmaSLCMaster ** 2) \
+    ds_out['IntensityAvgMaster'] = (np.abs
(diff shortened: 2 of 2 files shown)


ACHMartin:
bravo! For a review like this changing the heart of the code, could you please change the _version.py.
Ideally, it would be nice to have a unit_test on these fields + coherence to check our calculation are as expect.

With the OSCAR Med data, do you still have coherence values >1, did you check with the Iroise Sea data? We saw that in the initial OSCAR Iroise Sea data in my notebooks the coherence was always below 1. If it is not the case for the Iroise Sea data you have, it would be nice to dig why?

ACHMartin:
> bravo! For a review like this changing the heart of the code, could you please change the _version.py. Ideally, it would be nice to have a unit_test on these fields + coherence to check our calculation are as expect.
> 
> With the OSCAR Med data, do you still have coherence values >1, did you check with the Iroise Sea data? We saw that in the initial OSCAR Iroise Sea data in my notebooks the coherence was always below 1. If it is not the case for the Iroise Sea data you have, it would be nice to dig why?

In a perfect world you should run "pytest" before merging. Pytest is not passing one of the file (it is to be updated), but it is a good habit. In your case, I don't think the field your are changing are tested, so it will not change anything. But it will be good to test them.
I am quite new with unit_test, and I don't think I implemented the cleverer way. Probably Javier can help and give us some recommandations.

DavidMcCann-NOC (approved):
Well done for finding this bug - glad it was something simple

PR #273, opened 2023-07-27
Sorted the os.listdir command

elemerle:
Hi guys, again, 

I am unstoppable now :)

I added the function"sorted" to the os.listdir command because on my computer when I did a listdir it gave me the files in a random order which was not good for the find-triplets function, That is why I am asking this modification.

Eva


ACHMartin:
@DavidMcCann-NOC probably worth checking it doesn't kill your processing chain on any of your computer

DavidMcCann-NOC (approved):
Looks good!

PR #286, opened 2023-10-03
Update for readthedocs

DavidMcCann-NOC:
Small updates to the RTD config to get it to work after a change in how RTD operates.


ACHMartin (approved):
perfect. Before merging it, could you please update the __version.py file

PR #289, opened 2024-02-23
Add spatial ambiguity removal to main

JWMichalski:


seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Returns
+    -------
+    selected_ambiguity: ``int``
+        Index of the selected ambiguity
+    """
+    radius = int((box_size-1)/2)
+    L2_sel = lmout.isel(CrossRange=i_x, GroundRange=i_y)
+    if not np.isnan(L2_sel.isel(Ambiguities=0).CurrentU.values):
+        total_cost = cost_function(
```
ACHMartin, line 148: I am not in favour to call directly a function, I would prefer to get the output from the function called.

I term of naming, I think distance or env_distance or geophysical_distance will be more specific, hence easier not to mix with anything else.

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
...
+                CrossRange=slice(i_x-radius, i_x+radius+1),
+                GroundRange=slice(i_y-radius, i_y+radius+1)),
+            weight)
+        selected_ambiguity = total_cost.argmin()
+    else:
+        selected_ambiguity = np.nan
+    return selected_ambiguity
+
+
+def solve_ambiguity_spatial_selection(lmout, initial, cost_function,
```
ACHMartin, line 159: the default values are defined elsewhere, it shouldn't be the case again here.

If we want to solve the ambiguity only based on current or current, is there a way? We can set the ratio to 0, but not necessary to infinity. Should we use the same keyword as in "solve_ambiguity" with "optional `method` within `windcurrent` (default), `wind`, `current`".

method is probably not a very good keyword, but if we go this route, I suggest to open an issue and to change it in a next Pull Request.

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
...
+            }] = lmout.isel(
+                CrossRange=i,
+                GroundRange=j,
+                Ambiguities=selected_ambiguity
+            )
+
+    # initialize arrays
+    cross_range_size = lmout.CrossRange.sizes['CrossRange']
+    ground_range_size = lmout.GroundRange.sizes['GroundRange']
+    halfway_ground_range = int(ground_range_size/2)
```
ACHMartin, line 218: use round instead?

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
...
+
+    for n in range(pass_number):  # repeat passes
+        print('Pass', n+1)
+        # Pass A: iterate vertically
+        j = halfway_ground_range
+        while j >= 0:  # iterate across track
```
ACHMartin, line 224: I believe you could use:
https://docs.xarray.dev/en/stable/generated/xarray.DataArray.rolling.html

It will certainly be cleaner as a code and quicker.

I don't think we will be able to start from the centre, I don't find information on the windows_kwargs.

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,124 @@
+"""Testing of spatial_ambiguity_selection"""
+
+import pytest
+import numpy as np
+import xarray as xr
+import seastar.retrieval.spatial_ambiguity_selection as spatial_ambiguity_selection
+
+
+@pytest.fixture
+def L2():
```
ACHMartin, line 10: change to L2_2D?

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,124 @@
...
+    """Create a sample L2 OSCAR dataset"""
+    values = np.array([[3, 3, 3], [3, 3, 3], [3, 3, 3]])
```
ACHMartin, line 12: what are the dimensions? CrossRange, Groundrange and ?

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def test_Euclidian_distance(L2, lmout_single):
+    """Test the Euclidian distance cost function"""
+    total_cost = spatial_ambiguity_selection.Euclidian_distance(
+        lmout_single, L2, 1)
+    assert (total_cost == [54., 18., 0., 18.]).all()
+
+
+def cost(L2_sel, L2_neighbours, weight):
```
ACHMartin, line 105: why not using the defined cost function the Euclidian...? Or their future functions?

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,124 @@
...
+            ),
+            "EarthRelativeWindV": xr.DataArray(
+                values, dims=("CrossRange", "GroundRange")
+            ),
+        }
+    )
+
+
+@pytest.fixture
+def lmout_single():
```
ACHMartin, line 32: lmout_single, might be a .sel of lmout?

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
ACHMartin: L2_small2D_noisy?

ACHMartin (commented):
Thanks Kuba for this PR. I have made plenty of comments, sorry about that, but as a first python/xarray/github document it is really good.

My main comment is: I believe most of the loops could be changed as matrix calculation.
by starting with: https://docs.xarray.dev/en/stable/generated/xarray.DataArray.rolling.html
Could you please have a try, and we can discuss this on github directly.

Did you run pytest? Did it pass? We will need to review the pytest, but very good that you started. Thanks.

On a more practical point of view. What matters? The window size? Where does the ambiguity removal start with (centre, edge)? The number of iteration?
Thanks

seastar/retrieval/ambiguity_removal.py
```diff
@@ -111,13 +114,20 @@ def solve_ambiguity(lmout, ambiguity):
             `truth` HAVE to be in the dict.
             optional `method` within `windcurrent` (default), `wind`, `current`
             optional `windcurrentratio` default = 10
+        - name = `spatial_selection`
+            `costfunction` HAVE to be in the dict
+            `initial solution` HAVE to be in the dict
+            optional 'windcurrentratio' default = 5
```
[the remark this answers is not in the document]
JWMichalski: For now, I will just change the default to 10, I should've done this in the first place

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
...
+    box_size : ``int``, optional
+        Size of the box around the cell
+        Default is 3
+
+    Returns
+    -------
+    selected_ambiguity: ``int``
+        Index of the selected ambiguity
+    """
+    radius = int((box_size-1)/2)
```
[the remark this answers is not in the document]
JWMichalski: np.round() rounds half-integers to the nearest even integer, which is not desired here.
I will use np.floor

seastar/retrieval/spatial_ambiguity_selection.py
JWMichalski: I am not entirely sure what you mean here with calling a function directly instead of an output?

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
JWMichalski: This way if Euclidian distance fails the test, the other functions still pass.
As cost function is replacable with anything and not hardcoded into the function, I didn't want the other functions to fail the test if Euclidian fails.

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
+import numpy as np
+import xarray as xr
+
+
+def squared_Euclidian_distance(L2_sel, L2_neighbours, weight):
```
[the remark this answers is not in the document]
JWMichalski: I initially attempted to do this without loops.
Using xarray's tools, would, I believe, not overwrite the ambiguities 'as it goes', but it would do it all at once which gives significantly worse results (I have tested this using loops: the first version of my code was only selecting ambiguities at the end, after iterating through all of the track).

JWMichalski:
Currently seastar.retrieval.spatial_ambiguity_selection.solve_ambiguity_spatial_selection passes or **kwargs directly to cost_function.
This means that when somebody calls this function and makes a typo in a keyword argument, it will pass this argument to cost_function and the error will indicate that cost_function got an unexpected keyword argument. This is a very confusing behaviour for user and makes debugging difficult.
There are a few ways to solve this:
1.  require cost_function to take **kwargs. Not ideal, because then in a case of a typo there will be no checking and it will silently select a default value
2. check whether **kwargs match arguments expected by cost_function before calling, it and raise an error with a better explanation of what is going on
3. check whether **kwargs match arguments expected by cost_function before calling it, remove unexpected arguments and raise a warning instead of stopping the script

ACHMartin:
> Currently seastar.retrieval.spatial_ambiguity_selection.solve_ambiguity_spatial_selection passes **kwargs directly to cost_function. This means that when somebody calls this function and makes a typo in a keyword argument, it will pass this argument to cost_function and the error will indicate that cost_function got an unexpected keyword argument. This is a very confusing behaviour for user and makes debugging difficult. There are a few ways to solve this:
> 
>     1. require cost_function to take **kwargs. Not ideal, because then in a case of a typo there will be no checking and it will silently select a default value
> 
>     2. check whether **kwargs match arguments expected by cost_function before calling, it and raise an error with a better explanation of what is going on
> 
>     3. check whether **kwargs match arguments expected by cost_function before calling it, remove unexpected arguments and raise a warning instead of stopping the script

Could you please indicate directly in the code or refer to the lines. I understand your point, but I am not sure it is a big issue. I feel it is the usual behavior. Perhaps you could try with matplotlib.plot to see how it behaves. Otherwise, I tend to prefer to check the attributes within the function, for exemple at the early beginning of the function and raise issues if needed.
I feel it strange to test it before the function is called, as the function might be called from different places. I hope it make sense.

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,251 @@
...
+    -------
+    selected_ambiguity: ``int``
+        Index of the selected ambiguity
+    """
+    if window % 2 == 0:
+        raise ValueError('Window size must be an odd number')
+    radius = np.int_((window-1)/2)
+    L2_sel = lmout.isel(CrossRange=i_x, GroundRange=i_y)
+    if not np.isnan(L2_sel.isel(Ambiguities=0).CurrentU.values):
+        total_cost = cost_function(
```
JWMichalski, line 133: Here the cost_function is called and given kwargs as arguments, which will give an error if any of keys in kwargs are not arguments of cost_function

JWMichalski:
> > Currently seastar.retrieval.spatial_ambiguity_selection.solve_ambiguity_spatial_selection passes **kwargs directly to cost_function. This means that when somebody calls this function and makes a typo in a keyword argument, it will pass this argument to cost_function and the error will indicate that cost_function got an unexpected keyword argument. This is a very confusing behaviour for user and makes debugging difficult. There are a few ways to solve this:
> > ```
> > 1. require cost_function to take **kwargs. Not ideal, because then in a case of a typo there will be no checking and it will silently select a default value
> > 
> > 2. check whether **kwargs match arguments expected by cost_function before calling, it and raise an error with a better explanation of what is going on
> > 
> > 3. check whether **kwargs match arguments expected by cost_function before calling it, remove unexpected arguments and raise a warning instead of stopping the script
> > ```
> 
> Could you please indicate directly in the code or refer to the lines. I understand your point, but I am not sure it is a big issue. I feel it is the usual behavior. Perhaps you could try with matplotlib.plot to see how it behaves. Otherwise, I tend to prefer to check the attributes within the function, for exemple at the early beginning of the function and raise issues if needed. I feel it strange to test it before the function is called, as the function might be called from different places. I hope it make sense.

I've checked it with matplotlib.plot and it does look like it behaves the same way as solve_ambiguity_spatial_selection now, i.e. passes all **kwargs down into other functions (Line2D.set in this case) and returns got an unexpected keyword argument.
You're right, it does then seem to be the usual behaviour, so in this case it should stay as it is

seastar/retrieval/spatial_ambiguity_selection.py
JWMichalski: Starting from the centre is fairly important due to larger noise on the edges

seastar/retrieval/spatial_ambiguity_selection.py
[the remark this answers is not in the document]
JWMichalski: Rewritten using xarray's tools

seastar/retrieval/spatial_ambiguity_selection.py
```diff
@@ -0,0 +1,237 @@
...
+                        L2_sel.sel(
+                            Ambiguities=iambiguity).EarthRelativeWindV.values
+                        - L2_comp_neighbour.EarthRelativeWindV.values)**2
+                    total_cost[iambiguity] += weight*current_distance + \
+                        wind_distance
+    total_cost = xr.where(np.isnan(total_cost), np.inf, total_cost)
+    return total_cost
+
+
+def Euclidian_distance(L2_sel, L2_neighbours, weight):
```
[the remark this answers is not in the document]
JWMichalski: Merged into one function now

seastar/retrieval/ambiguity_removal.py
```diff
@@ -138,10 +149,29 @@ def solve_ambiguity(lmout, ambiguity):
                                                 ambiguity['truth'],
                                                 windcurrentratio=ambiguity['windcurrentratio'])
         sol = lmout.isel(Ambiguities=index_dict[ambiguity['method']])
+    elif ambiguity['name'] == 'spatial_selection':
+        if 'costfunction' not in ambiguity:
+            raise Exception(
+                "ambiguity['costfunction'] HAVE to be provided for spatial_selection method")
```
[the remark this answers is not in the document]
JWMichalski: Not needed anymore; Euclidian distance is called directly

seastar/retrieval/spatial_ambiguity_selection.py
JWMichalski: Euclidian distance called directly now, instead of being passed as cost_function

seastar/retrieval/spatial_ambiguity_selection.py
JWMichalski: keywords uniform with other functions now

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
JWMichalski: Renamed

seastar/tests/retrieval/test_spatial_ambiguity_selection.py
JWMichalski: Outdated, now the cost_function is hardcoded so no need to define it here

seastar/retrieval/spatial_ambiguity_selection.py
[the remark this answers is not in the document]
JWMichalski: Additionally rounding using np.int_(), this is the fastest option to take floor as integer

Issue #290, opened 2024-04-18
String concatanation used instead of os.path.join for filepaths merging in seastar.oscar.tools.load_OSCAR_data


JWMichalski:
String concatanation of filepaths can cause issues with slashes.
Solution:
Replace
```
ds[ind] = readNetCDFFile(file_path + file_list[ind])
```
by
```
ds[ind] = readNetCDFFile(os.path.join(file_path, file_list[ind]))
```

PR #291, opened 2024-04-18
Issue #290

JWMichalski:
Resolve issue #290


seastar/.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
+{
```
DavidMcCann-NOC, line 1: What is this .json file? Its not something i've come across before - best to exclude it from this pull request if it isn't anything to do with the issue (i assume it slipped itself in!).

DavidMcCann-NOC (changes requested):
Looks good to me. If you could remove the settings.json change from the commit (if it isn't necessary) then its all good to go.

seastar/.vscode/settings.json
DavidMcCann-NOC: > It is not necessary, it's settings for VScode. I have it in .gitignore in my branch and commited it here by accident. Maybe it would be good to add it to .gitignore in main?

Yes, perhaps so. However will it cause issues for you working on, e.g., new branches from /main that don't have it? Or is it something that gets automatically created by VScode anyway? If the latter then yes - good to add to .gitignore in /main. If the former, then i'm happy to let it stay

JWMichalski: This is something that VScode automatically creates every time settings are changed on a new branch.
In my case this happens because seastar uses pytest for testing, so VScode creates a settings file that has paths to tests and other pytest-related settings.

PR #293, opened 2024-04-18
add VScode settings.json to .gitignore

JWMichalski:
VScode creates a settings.json file every time settings are changed on a new branch. This is IDE-specific and doesn't need to be included in the repository - adding it to .gitignore avoids accidental commits


DavidMcCann-NOC (approved):
Great, thanks for that Kuba

Issue #298, opened 2025-01-31
Bad quality of intensity of Mid beam


elemerle:
Intensity images from the Mid beam have a bad quality and seem to have a coarser resolution.

PR #299, opened 2025-01-31
Correction of compute_multilooking_Master_Slave

elemerle:
Correction of compute_multilooking_Master_Slave in level1 oscar for the calculation of intensity for the Mid beam #298

### DIFF
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -275,45 +275,46 @@ def compute_multilooking_Master_Slave(ds, window=3,
     if len(ds.SigmaImageSingleLookRealPart.dims) > 2:
         raise Exception("The variable SigmaImageSingleLookRealPart is not a"
                         "2D variable. Please check this variable's dimensions")
-    ds_out = xr.Dataset()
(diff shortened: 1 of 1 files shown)


DavidMcCann-NOC (approved):
This looks good to me, and thanks for reorganising the code a bit too - you're right, its easier to read now.

seastar/oscar/level1.py
```diff
@@ -275,71 +275,71 @@ def compute_multilooking_Master_Slave(ds, window=3,
     if len(ds.SigmaImageSingleLookRealPart.dims) > 2:
         raise Exception("The variable SigmaImageSingleLookRealPart is not a"
                         "2D variable. Please check this variable's dimensions")
-    ds_out = xr.Dataset()
+    ds_out = ds.copy()
```
ACHMartin, line 278: why do we need a copy of the dataset?

ACHMartin (commented):
Before merging, you should change the file '_version.py' and update it to 2025.01 given it is a big change in the processing.

Please run pytest to check it works fine.

PR #305, opened 2025-02-13
Environment modif

ACHMartin:
Change of the pyproject.toml to have python = ">=3.8" instead of the previous python = "3.8.16" which was too strict for the dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ version = '1.0'
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python >= "3.8.16"                     # Python required version
+python = ">= 3.8"                     # Python required version
 matplotlib = "3.5.1"                      # Matplotlib
 netcdf4 = "1.5.7"                         # NetCDF4
 notebook = "6.4.11"                       # Notebook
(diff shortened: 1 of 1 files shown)


ACHMartin:
> This will be an interesting test, as I _think_ I had some dependency issues with later versions of Python, it will need testing for sure.
> 
> Also, if this is changed here we will also need to change the specified version in the yaml file in case there is a similar strictness

Personnaly, I believe we could be less strict about every package for the dependencies. Probably better to check this when we have proper "pytest".
In our case, the YML was not strict on python, it is 3.8. So in my latest installation I got 3.8.18, but the dependency was 3.8.16. Hence the issue.
For sure pyproject.toml HAS to be less strict than the YML in order not to raise this issue again.
It is the case now.

DavidMcCann-NOC:
> > This will be an interesting test, as I _think_ I had some dependency issues with later versions of Python, it will need testing for sure.
> > Also, if this is changed here we will also need to change the specified version in the yaml file in case there is a similar strictness
> 
> Personnaly, I believe we could be less strict about every package for the dependencies. Probably better to check this when we have proper "pytest". In our case, the YML was not strict on python, it is 3.8. So in my latest installation I got 3.8.18, but the dependency was 3.8.16. Hence the issue. For sure pyproject.toml HAS to be less strict than the YML in order not to raise this issue again. It is the case now.

Ah i see! Yes, very much agree

DavidMcCann-NOC (approved):

PR #310, opened 2025-02-18
Cmod7 addition

ACHMartin:
Addition of the CMOD7 function in the GMFs, modification of compute_nrcs to include cmod7 gmfs option.
Addition of pytest for cmod7. Pytest successful for test_nrcs (test_wind_current_retrieval  fail, as usually the case)
I didn't change the "_version_", as I assume this change to be minor.


elemerle:
Please double check the commit link cmod7 function into compute_nrcs. The part where you raise the exception.

elemerle (approved):

PR #316, opened 2025-03-19
Changes to L1AP processing

DavidMcCann-NOC:
Added calculation of attribute parameters and injected them into the L1AP file upon processing.

Script now copies the L1A data into an L1AP directory with the same branch structure as L1A (i.e., Campaign\L1AP\data_version\flight\resolution)

Reads track name from an ini dictionary now included in the matlab scripts folder.

Also included an INI from mathworks to read this file

Still plenty of improvement can be made, e.g. adding more attributes, adding 'L1AP' to the save file name, changing the format of StartTime etc.


matlab/metasensing/track_names.ini
ACHMartin, line 1: I suggest we don't put the .ini in matlab/metasensing but higher up in the tree. Potentially in a dedicated "ini" directory at the same level as docs or seastar.

Should we have a dedicated track_names.ini per campaign eg IroiseSea2022_track_names.ini; MedSea2023_track_names.ini?
I think it might be more resilient to error as there is less risk that we changed things for campaing 2022 while processing campaign 2026 (hopefully :) )

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -22,7 +49,59 @@
...
+    end_hour = ncread([file_path, file_name],'FinalHour');
+    end_minute = ncread([file_path, file_name],'FinalMin');
+    end_time = sprintf('%02d%02d%02dT%02d%02d',[end_year,end_month,end_day, end_hour, end_minute]);
+
+    % Write global attributes
+    ncwriteatt([file_path, file_name], '/','Campaign', campaign_name)
+    ncwriteatt([file_path, file_name], '/','Platform', 'OSCAR')
+    ncwriteatt([file_path, file_name], '/','ProcessingLevel', 'L1AP')
+    ncwriteatt([file_path, file_name], '/','Track',track_name)
+    ncwriteatt([file_path, file_name], '/','StartTime',start_time)
```
ACHMartin, line 96: I guess it is difficult to get start time in datetime with Matlab?

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -22,7 +49,59 @@
...
+    disp(['Processing file ',file_name])
+
+    % Read track_names.ini to match Track name from DAR to track time
+    track_time_for_ini_search = ['x',erase(ncreadatt([file_path, file_name],"/","Title"), 'Track : ')];
+    ini_file = INI('File','track_names.ini');
+    ini_struct = ini_file.read();
+    track_name = ini_struct.(track_time_for_ini_search(1:9)).(track_time_for_ini_search);
+
+    % Read Python _version.py to retrieve __version__ parameter
+    text = fileread('../../_version.py ');
```
ACHMartin, line 61: this paragraph on getting the "version" seems very complicated! I guess it is because it has been formated for python and not matlab

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -8,6 +8,33 @@
...
+    save_path_components = file_path_components;
+    save_path_components{find(strcmp(file_path_components, 'L1A'))} = 'L1AP';
+    if strcmp(OS, 'WINDOWS')
+        save_path = cell2mat(join(save_path_components,'\'));
+        file_path_split = split(file_path,'\');
+    elseif strcmp(OS, 'UNIX')
+        save_path = cell2mat(join(save_path_components,'/'));
+        file_path_split = split(file_path,'/');
+    end
+    % copy L1A file to L1P directory to be worked on
```
ACHMartin, line 29: in comments, change L1P to L1AP

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
```diff
@@ -8,6 +8,33 @@
...
+    if ~exist(save_path, 'dir')
+        mkdir(save_path)
+    end
+    copyfile([file_path, file_name], [save_path, file_name], 'f')
```
ACHMartin, line 33: I don't understand why we copy the L1A files in the L1AP folder

ACHMartin: Ok I think I understand, the function is adding/creating variables/attributes to the existing netCDF

ACHMartin (approved):
I think the only main question in my comments is do we have different track_names.ini files for different campaigns

matlab/metasensing/track_names.ini
DavidMcCann-NOC: > I suggest we don't put the .ini in matlab/metasensing but higher up in the tree. Potentially in a dedicated "ini" directory at the same level as docs or seastar.

Sure thing - this was an easy start, but we for sure can pop the ini files anywhere else.

> Should we have a dedicated track_names.ini per campaign eg IroiseSea2022_track_names.ini; MedSea2023_track_names.ini?
> I think it might be more resilient to error as there is less risk that we changed things for campaing 2022 while processing campaign 2026 (hopefully :) )

Yeah thats a good call, and much easier to have different files for different campaigns than one giant ini file with 3 layers instead of 2.

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: > I guess it is difficult to get start time in datetime with Matlab?

I can do that - its just a question of what format we want it in. I can format the start time as anything. Reading it from the L1A files requires those multiple calls to `ncread` because of how Matlab treats NetCDF files and how Karlus originally formatted the start and end times as separated variables in the L1A data

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: > this paragraph on getting the "version" seems very complicated! I guess it is because it has been formated for python and not matlab

Don't get me started.....

It looks horrendous but this -should- be robust to any changes to the `_version.py` file. This finds the __version__ variable and then extracts the string after it. But its doing it at a very low I/O level here. No `from _version import __version__` in matlab!

matlab/metasensing/add_inc_and_squint_to_netcdf_batch.m
DavidMcCann-NOC: > Ok I think I understand, the function is adding/creating variables/attributes to the existing netCDF

Yeah thats it. No easy way to create a dataset in memory and then export it to NetCDF like with xarray. This is cumbersome - but it means that we don't work on the L1A files and change them, which is how Karlus' script worked initially.

matlab/metasensing/track_names.ini
DavidMcCann-NOC: Should we call the folder 'config' and have all of the configuration files in there?

ACHMartin: 'config' looks fine

PR #319, opened 2025-03-20
New function to create short name from MD5 checksum

DavidMcCann-NOC:
New functions to create MD5 checksum in hexidecimal string format and a (very) simple function to create a short name based on an MD5 checksum.


ACHMartin (approved):
perfect, thanks

PR #322, opened 2025-03-24
read_OSCAR_track_names_config

DavidMcCann-NOC:
New function to read OSCAR track names config using campaign name as input.

Apologies for the confusion in the last pull request - I've cleaned up the commits.


DavidMcCann-NOC:
To use this function at the begining of the L1B processing chain, for example:


```
campaign = '202205_IroiseSea'
processing_level = 'L1AP'
version = 'v20250312'
flight = '20220522'
resolution = '008x008m'

track_config = seastar.utils.readers.read_OSCAR_track_names_config(campaign, flight)
L1AP_path = os.path.join(path_to_data, campaign, processing_level, version, flight, resolution)

file_time_triplets = seastar.oscar.tools.find_file_triplets(L1AP_path)
DAR_tracks=dict()
DAR_tracks[flight] = {track_config[file_time_triplets[i][0]]:i for i in range(len(file_time_triplets))}
```
This will make a dict of {Track_name : File_index} for looping through in processing

elemerle:
Thanks David,

Do we make the resolution appear (in path and attrs) from the L1AP or from the L1B during the processing? 

Because I do not see the resolution in the LA1P attributes? And does it make sens to mention resolution for L1AP ? Will it be modified?

DavidMcCann-NOC:
> Thanks David,
> 
> Do we make the resolution appear (in path and attrs) from the L1AP or from the L1B during the processing?
> 
> Because I do not see the resolution in the LA1P attributes? And does it make sens to mention resolution for L1AP ? Will it be modified?

Ah, good point. 

> Thanks David,
> 
> Do we make the resolution appear (in path and attrs) from the L1AP or from the L1B during the processing?
> 
> Because I do not see the resolution in the LA1P attributes? And does it make sens to mention resolution for L1AP ? Will it be modified?

For the path, this is set from the very beginning - the L1A data should be saved in (as per the file format doc):
`\campaign\processing_level\version\flight_date\resolution\` (with resolution being 008x008m in the case of MetaSensing L1A data)
The L1AP script automatically mirrors this structure if it doesn't already exist and creates a new L1AP tree under `\campaign`.

For the attributes, this needs to be added in at L1AP I think, as its the first time we are adding any attributes ourselves to the NetCDF files. I'll add an issue - this will need to be added to the new `L1A_to_L1AP_processing()` function i added this morning too, so best to change in a different pull request.

ACHMartin:
> For the path, this is set from the very beginning - the L1A data should be saved in (as per the file format doc): `\campaign\processing_level\version\flight_date\resolution\` (with resolution being 008x008m in the case of MetaSensing L1A data) The L1AP script automatically mirrors this structure if it doesn't already exist and creates a new L1AP tree under `\campaign`.
> 

Perhaps for L1A and L1AP, given we keep the raw resolution perhaps the "resolution" folder is meaningless.

> For the attributes, this needs to be added in at L1AP I think, as its the first time we are adding any attributes ourselves to the NetCDF files. I'll add an issue - this will need to be added to the new `L1A_to_L1AP_processing()` function i added this morning too, so best to change in a different pull request.

Not sure we need the resolution for the L1AP. @elemerle did a function to calculate "resolution" for the L1B, so perhaps we can have resolution only from there, when we start to play with it.

ACHMartin (approved):
good for the new code in the pull request.

DavidMcCann-NOC:
> > For the path, this is set from the very beginning - the L1A data should be saved in (as per the file format doc): `\campaign\processing_level\version\flight_date\resolution\` (with resolution being 008x008m in the case of MetaSensing L1A data) The L1AP script automatically mirrors this structure if it doesn't already exist and creates a new L1AP tree under `\campaign`.
> 
> Perhaps for L1A and L1AP, given we keep the raw resolution perhaps the "resolution" folder is meaningless.
> 
> > For the attributes, this needs to be added in at L1AP I think, as its the first time we are adding any attributes ourselves to the NetCDF files. I'll add an issue - this will need to be added to the new `L1A_to_L1AP_processing()` function i added this morning too, so best to change in a different pull request.
> 
> Not sure we need the resolution for the L1AP. @elemerle did a function to calculate "resolution" for the L1B, so perhaps we can have resolution only from there, when we start to play with it.

It would be easier if all data had a `/resolution` directory so that it standardises the path structure across all OSCAR data file types. We're currently using `split(file_dir, '/')' to deconstruct the directory, so it makes things smoother if everything ends up with a resolution field. I agree its extraneous, given that the resolution of L1A and L1AP data will likely never change for OSCAR, but if we scrub that directory level below L1B we'll have to be careful

Issue #333, opened 2025-03-28
New L1AP file names have broken ss.oscar.tools.find_file_triplets()


DavidMcCann-NOC:
`ss.oscar.tools.find_file_triplets` Can't read the new L1AP files, likely because of the inclusion of `'L1AP_'` at the beginning of the L1A file names. The output is no longer the file triplets dict, but ,e.g.:
```
[('13.5G',
  [0,
   1,
   2,
...
   44])]
```
Which points to an outdated means of reading the file names.

elemerle:
Strange, this function works for me in the previous merge because I have taken into account the inclusion of 'L1AP_' at the beginning of the L1A file names: cf line 100 in new  seastar/oscar/tools.py

Anyway I think we have to change this function to read the date from the config file

DavidMcCann-NOC:
> Strange, this function works for me in the previous merge because I have taken into account the inclusion of 'L1AP_' at the beginning of the L1A file names: cf line 100 in new seastar/oscar/tools.py
> 
> Anyway I think we have to change this function to read the date from the config file

It might be a bug in how i was using it at my end then - I'll investigate later today, its not as critical if its working for you at the moment.

There might be a better way of organising the function, but good idea about linking back to the campaign config files as it may well simplify things.

DavidMcCann-NOC:
Closed with #334

PR #334, opened 2025-03-28
Update to `seastar.utils.tools.find_file_triplets` to address Issue #333

DavidMcCann-NOC:
More robust implementation of `find_file_triplets()` that utilises regular expressions to find the track time of every file in a folder. Previous implementation relied on the time being in an explicit position in `sr.split(file_name, '_')` which caused Issue #333.


DavidMcCann-NOC:
No idea why there are 4 empty commits along with this one, something to do with me updating my test branch from main I think. There seems to be only one file change confirmed in the pull request though

ACHMartin (approved):
it seems alright

elemerle:
Hello,
We discuss a bit with @ACHMartin about the processing function and we conclude on the fact that it's better if the processing runs for one track individually.
So, I have started to code the processing_L1APtoL1B. And I do not use the find_file_triplets function anymore since with the track dict we have access to the date for each track to find the triplet file I do this:

    # Getting the date for every tracks in the dict.
    track_names_dict = ss.utils.readers.read_OSCAR_track_names_config(campaign, acq_date)
    
    # Getting the date for the track we are interested in
    if track in track_names_dict.values():
        date_of_track = [key for key, v in track_names_dict.items() if v == track][0]
        logger.info(f"Track '{track}' found, acquisition time: {date_of_track}")
    else:
        logger.warning(f"Track '{track}' not found in track_names_dict. The code will crash.")
        
    # Loading the file names of the files corresponding to date_of_track (triplet of file - one per antenna)
    L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_path + "/*" + date_of_track + "*.nc"))]

elemerle:
This method provide a list of the file name sorted so that the antenna indent works and we do not have the indexes which I found a bit confusing before.

I will be happy to discuss this with both of you to converge through a solution that suits for everybody.

DavidMcCann-NOC:
> This method provide a list of the file name sorted so that the antenna indent works and we do not have the indexes which I found a bit confusing before.
> 
> I will be happy to discuss this with both of you to converge through a solution that suits for everybody.

Yeah this is a better way to go about finding the triplets - thanks Eva. I agree, the file indices are a bit crude when we know and have already defined exactly what time each track is for each flight. Much clearer

DavidMcCann-NOC:
I'm going to merge this now, and then we can decide on the redundancy of the function after Eva's changes.

PR #335, opened 2025-03-28
Modification of reading file

elemerle:
Get the file list with the date.


seastar/oscar/tools.py
```diff
@@ -35,10 +34,9 @@ def load_OSCAR_data(file_path, file_inds):
         keys as ``int`` to identify each antenna
 
     """
-    file_list = sorted(os.listdir(file_path))
     ds = dict()
-    for ind in file_inds:
-        ds[ind] = readNetCDFFile(os.path.join(file_path, file_list[ind]))
+    for ind in range(len(file_list)):
```
DavidMcCann-NOC, line 38: Could probably use enumerate there to be ever so slightly more pythonic, e.g.:
```
for ind, file_name in enumerate(file_list):
    ds[ind] = readNetCDFFile(os.path.join(file_path, file_name))
```

seastar/oscar/tools.py
```diff
@@ -12,21 +12,20 @@
...
     `{file index : xarray.Dataset}`
 
     Parameters
     ----------
     file_path : ``str``
         Path containing OSCAR data files
-    file_inds : ``list``
-        List of indices of the Fore/Mid/Aft antenna data triplet within the
-        file list generated by `os.listdir(file_path)`
+    file_list : ``list``
```
DavidMcCann-NOC, line 27: Probably best to be explicit about what `file_list` is, i.e. a ``list`` of ``str``

seastar/oscar/tools.py
```diff
@@ -149,10 +144,9 @@ def identify_antenna_location_from_filename(file_path, file_time_triplets):
 
     """
     antenna_identifiers = {'0': 'Mid', '3': 'Fore', '7': 'Aft'}
-    file_list = sorted(os.listdir(file_path))
     antenna_id = list()
-    for i in range(len(file_time_triplets)):
-        file_name = file_list[file_time_triplets[i]]
+    for i in range(len(file_list)):
```
DavidMcCann-NOC, line 148: Is the i needed? I think you can get away with something like:

```
for file name in file_list:
    antenna_id.append(antenna_identifiers[file_name.split('_')[6][0]])
```

ACHMartin (approved):
some comments on naming and documentation

seastar/oscar/tools.py
ACHMartin: you're becoming very pythonic ;)

seastar/oscar/tools.py
DavidMcCann-NOC: It also might be better to explicitly search for the antenna ID in the file name string, just in case that the 6th index in the split file name is something else. Unlikely in the short term, but I've already proved that I can break my own function by changing the filenames slightly...

A regular expression can be used to find the entry in the split name that is exactly two repeated digits (e.g., 00, 33, etc). I'll just figure out an example for you

seastar/oscar/tools.py
```diff
@@ -35,10 +34,9 @@ def load_OSCAR_data(file_path, file_inds):
         keys as ``int`` to identify each antenna
```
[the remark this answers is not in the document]
DavidMcCann-NOC: Isn't it because the entries are xr.Datasets of different size (i.e., mid beam is smaller with fewer variables). You can only have the same sized datasets in a list, I think - which is probably why i used dicts in the first place when grouping the three beam datasets together for processing. Functions later in the chain unify the datasets and then merge them together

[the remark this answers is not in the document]
elemerle: I do not use the triplet function anymore since I use this line (in a hidden code) to get the L1AP file names: L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_path + "/*" + date_of_track + "*.nc"))].

I find the indexes a bit confusing so I think maybe we can delete this function?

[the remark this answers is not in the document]
DavidMcCann-NOC: > I do not use the triplet function anymore since I use this line (in a hidden code) to get the L1AP file names: L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_path + "/_" + date_of_track + "_.nc"))].
> 
> I find the indexes a bit confusing so I think maybe we can delete this function?

Lets get your new processing chain and methodology working first, then we can clean up any unused functions as a separate issue. Just in case we want to go back and revisit them for whatever reason

seastar/oscar/tools.py
```diff
@@ -12,21 +12,20 @@
...
 
-def load_OSCAR_data(file_path, file_inds):
+def load_OSCAR_data(file_path, file_list):
     """
     Load OSCAR data.
 
     Loads in OSCAR Fore/Mid/Aft antenna files from the path designated in
-    `file_path` and file indices in `file_inds`. Returns data in a dict of
+    `file_path` and files in `file_list`. Returns data in a dict of
     `{file index : xarray.Dataset}`
```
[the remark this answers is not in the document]
elemerle: Maybe load_L1AP_oscar_data is more explicit. The function returns a dict of dataset with the 3 xr.datasets which correspond to one dataset per antenna

seastar/oscar/tools.py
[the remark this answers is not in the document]
elemerle: Yeah it's okay for me, Should I share my processing function in a new merge request or in this one?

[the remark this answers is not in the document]
ACHMartin: for readibility, probably better in a new merge request

seastar/oscar/tools.py
DavidMcCann-NOC: To find the antenna identification in the L1A / L1AP file name using regular expressions:

```
import re
r = re.compile(r'^(?!.*[a-zA-Z])(?:\D*\d\D*\d\D*)$')
antenna_filter = filter(r.match, file_list[0].split('_'))
antenna_number = "".join(map(str, antenna_filter))
antenna_identifiers = {'0': 'Mid', '3': 'Fore', '7': 'Aft'}
antenna = "".join({antenna_identifiers[i] for i in antenna_number if i in antenna_identifiers})
```
This will split the filename using the '_' delimiter, and then hunt in the resulting list using the regex for any entry that does not contain any letters and is exactly two digits long. The filter its then turned into a number, which we then turn into an antenna identifier using a dictionary, matching to the antenna name if any of the digits in antenna_number match the dict keys.

Slightly long-winded, but from memory the antenna identifier in the file name does not have to be exactly '00, 33 or 77', but if 3 or 7 are in the identifier it is always fore or aft.

DavidMcCann-NOC:
```
r = re.compile(r'^(?!.*[a-zA-Z])(?:\D*\d\D*\d\D*)$')
antenna_filter = filter(r.match, file_name.split('_'))
antenna_number = "".join(map(str, antenna_filter))
antenna_identifiers = {'0': 'Mid', '3': 'Fore', '7': 'Aft'}
antenna = "".join({antenna_identifiers[i] for i in antenna_number if i in antenna_identifiers})
```

This will split the filename using the '_' delimiter, and then hunt in the resulting list using the regex for any entry that does not contain any letters and is exactly two digits long. The filter its then turned into a number, which we then turn into an antenna identifier using a dictionary, matching to the antenna name if any of the digits in antenna_number match the dict keys.

Slightly long-winded, but from memory the antenna identifier in the file name does not have to be exactly '00, 33 or 77', but if 3 or 7 are in the identifier it is always fore or aft.

DavidMcCann-NOC: Oh and you'll need to `import re` if you don't already

DavidMcCann-NOC: Even simpler regex:

```
r = re.compile(r'(?:\D*\d\D*\d\D*)$')

```

Just looks for exactly two digits.

elemerle: Okay thhanks Dave, I'll implement that :)

elemerle:
I don't know what is going on with this branch it's a real mess to handle

DavidMcCann-NOC:
> I don't know what is going on with this branch it's a real mess to handle

Did you get a lot of conflicts?

elemerle:
I don't know why it merges during the push force. Any way this is merged, if there are issues please open new issues and I will handle it in a new branch

Issue #330, opened 2025-03-26
issue about the `units` attribute.


elemerle:
A small comment, but there might be a discrepancy across functions in how we define the `units` attribute. In the 'old' functions its, e.g.:
ds_wa.attrs['units'] = 'm/s'

rather than:
ds_wa.attrs['units'] = ['m/s']

MetaSensing have their units with square brackets, so perhaps we should standardise with that? (In which case I can go back and change the old functions). However personally i don't like the default of the square brackets on plots - its just an aesthetic thing.

_Originally posted by @DavidMcCann-NOC in https://github.com/ACHMartin/seastar_project/pull/314#discussion_r2007300889_

ACHMartin:
In xarray documentation (https://docs.xarray.dev/en/latest/user-guide/data-structures.html) they don't put any bracket for "units".
So I suggest to have: `ds_wa.attrs['units'] = 'm/s'`

elemerle:
From my side I do not have any preference as soon as we are consistent all along the code.

DavidMcCann-NOC:
> Maybe we have to modify the units from MetaSensing, I agree with David on the fact that brackets on plots is not very aesthetic.
> 
> If we modify the MetaSensing unit attribute, this should be done in the matlab processing no?

Yes, this would be an L1AP function in Matlab. Although it could be easier and faster to do it in Python, as playing around with NetCDF files is much easier here. 

We already have a function to check for missing variables, so we could have a simple function to look for units attributes that contain brackets and remove the brackets - I'll put something together to propose

PR #336, opened 2025-03-31
Update tools.py

DavidMcCann-NOC:
New function `clean_units_attribute` to address #330. Takes an `xr.Dataset` and finds and removes square brackets from all global and variable attributes named 'units'.

### DIFF
--- a/_version.py
+++ b/_version.py
@@ -7,5 +7,5 @@ Version of the codebase
 """
 
 
-__version__ = '2025.03.3'
+__version__ = '2025.03.4'
 
--- a/seastar/oscar/tools.py
+++ b/seastar/oscar/tools.py
@@ -305,3 +305,36 @@ def check_attrs_dataset(ds):
 
     return ds
 
+def clean_units_attribute(ds):
+    """
+    Clean units attributes from dataset.
+    
+    Looks for global and varible `units` attributes and removes all instances
+    of square brackets around the `units` value.
+    
+    e.g., '[m]' --> 'm'
+    
+    Parameters:
+        ds : ``xr.Dataset``
+            Input dataset to clean
+    
+    Returns:
+        ds_out : ``xr.dataset``
+            Dataset with cleaned units attributes.
+    """
+    
+    def remove_brackets(unit_str):
+        return re.sub(r'[\[\]]', '', unit_str)
+    
+    ds_out = ds.copy()
+    
+    # Check global attributes
+    if 'units' in ds_out.attrs and isinstance(ds_out.attrs['units'], str):
+        ds_out.attrs['units'] = remove_brackets(ds_out.attrs['units'])
+    
+    # Check each variable's attributes
+    for var_name, var in ds_out.variables.items():
+        if 'units' in var.attrs and isinstance(var.attrs['units'], str):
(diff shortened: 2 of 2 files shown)


ACHMartin (approved):

_version.py
```diff
@@ -7,5 +7,5 @@
 """
 
 
-__version__ = '2025.03.3'
+__version__ = '2025.03.4'
```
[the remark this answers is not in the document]
DavidMcCann-NOC: I'm going to have to tape a message on my screen to make sure i remember!

seastar/oscar/tools.py
```diff
@@ -305,3 +305,38 @@ def check_attrs_dataset(ds):
...
+    Returns
+    -------
+    ds_out : ``xr.dataset``
+        Dataset with cleaned units attributes.
+    """
+    
+    def remove_brackets(unit_str):
+        return re.sub(r'[\[\]]', '', unit_str)
+    
+    ds_out = ds.copy()
```
elemerle, line 331: I know that @ACHMartin doesn't like ds.copy() so I've asked an alternative to chatgpt which gave me this:
def clean_units_attribute(ds):
    return ds.pipe(
        lambda d: d.assign(
            attrs={k: re.sub(r'[\[\]]', '', v) if k == "units" and isinstance(v, str) else v for k, v in d.attrs.items()},
            **{var: d[var].assign_attrs(
                {k: re.sub(r'[\[\]]', '', v) if k == "units" and isinstance(v, str) else v for k, v in d[var].attrs.items()}
            ) for var in d.variables}
        )
    )
    
    It's an optimise function to avoid copying large dataset

DavidMcCann-NOC: Thats true, it is an unnecessary copy command. Wouldn't simply working on the dataset directly work? That way the output on the front end can be a different dataset if desired, e.g. `ds_new = clean_units_attribute(ds)`, otherwise we assume we want to work on the same version on the dataset as is being input. The function doesn't change anything if it doesn't find brackets in the units attributes so I would presume this would be safe:



```
def clean_units_attribute(ds):

    def remove_brackets(unit_str):
        return re.sub(r'[\[\]]', '', unit_str)

    # Check global attributes
    if 'units' in ds_out.attrs and isinstance(ds.attrs['units'], str):
        ds.attrs['units'] = remove_brackets(ds.attrs['units'])
    
    # Check each variable's attributes
    for var_name, var in ds.variables.items():
        if 'units' in var.attrs and isinstance(var.attrs['units'], str):
            var.attrs['units'] = remove_brackets(var.attrs['units'])
    
    return ds
```
Apologies, Adrien approved the merge so I just went ahead, but i can open this back up again to make the change

elemerle: Yeah I guess the function is safe so we can work on the dataset directly. 
Our dataset are not that voluminous so I assume that the copy() won't take to much time but it can be an improvement to keep in mind. @ACHMartin any view about that?

ACHMartin: thanks @elemerle, indeed if we can do without `ds.copy()`, I guess it is better.
I assume we can change directly the "units" without doing any copy.
Changing "ds_out" by "ds" should work?

PR #340, opened 2025-03-31
Processing chain L1AP to L1B

elemerle:
Here is the processing chain from L1AP to L1B

Please give me your suggestions :)

Here is how I launch the function:

import seastar as ss
import os
import platform

from _logger import logger

def get_base_directory():
    """Return the base directory based on the operating system."""
    if platform.system() == "Windows":
        return "N:\\"
    elif platform.system() == "Linux" :
        return "/PROJETS/"
    else: 
        logger.warning(f"Unknown system name: {platform.system()}. Your path to data may be corrupted.")
        return None 

acq_date = "20230508"
data_version = "v20250312"
processing_level = "_L1AP_testcode"
campaign_dict = ss.utils.readers.read_campaign_config()
campaign = campaign_dict[acq_date[:6]]

path_oscar_L1AP = os.path.join(get_base_directory(),"1491-FE_SEASTARex_CCN_2024", "REALISATION", "Technique", "Data", "OSCAR", campaign, processing_level, data_version, acq_date)

track_list = ["Track_L1", "Track_1"]

for track in track_list:
    ds = ss.oscar.level1.processing_OSCAR_L1APtoL1B(path_oscar_L1AP, campaign, acq_date, track, write_nc=True)

### DIFF
--- a/_logger.py
+++ b/_logger.py
@@ -4,7 +4,7 @@ import argparse
 # Parse command-line arguments
 parser = argparse.ArgumentParser(description="Run the script with optional verbose mode.")
 parser.add_argument("-v", "--verbose", action="store_true", help="Enable verbose mode (INFO logs)")
-args = parser.parse_args()
+args, unknown = parser.parse_known_args()
 
 # Logger configuration
 logging.basicConfig(
--- a/_logger.py
+++ b/_logger.py
@@ -1,8 +1,14 @@
 import logging
+import argparse
+
+# Parse command-line arguments
+parser = argparse.ArgumentParser(description="Run the script with optional verbose mode.")
+parser.add_argument("-v", "--verbose", action="store_true", help="Enable verbose mode (INFO logs)")
+args = parser.parse_args()
 
 # Logger configuration
 logging.basicConfig(
-    level=logging.INFO,  # Default level (can be DEBUG, WARNING, ERROR)
+    level=logging.INFO if args.verbose else logging.WARNING,  # Set level based on --verbose
     format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
     handlers=[
         logging.StreamHandler(),  # Display logs in the console
@@ -12,6 +18,3 @@ logging.basicConfig(
 
 # Create the logger
 logger = logging.getLogger("seastar")  # Logger name based on your package
-
-# Optional: Adjust the level depending on the environment
-logger.setLevel(logging.DEBUG)  # Enable in development mode
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -10,6 +10,7 @@ import warnings
 from datetime import datetime as dt
 from datetime import timezone
 
+from _version import __version__
 from _logger import logger
 
 
@@ -226,7 +227,10 @@ def add_central_electromagnetic_wavenumber(ds):
     return ds
 
 
-def compute_multilooking_Master_Slave(ds, dict_L1B_process):
+def compute_multilooking_Master_Slave(ds, window=3,
+                                      vars_to_send=['Intensity',
+                                                    'Interferogram',
+                                                    'Coherence']):
     """
     Compute  multilooking Master/Slave L1b image products.
 
@@ -243,11 +247,10 @@ def compute_multilooking_Master_Slave(ds, dict_L1B_process):
     ----------
     ds : ``xarray.Dataset``
         OSCAR L1a dataset
-    dict_L1B_process : ``dict``
-        Dictionnary containing information about the window for the rolling mean for the multilooking computation
-        and the vars to provide (vars_to_send) after the multilooking computation.
-        window : Integer averaging window size. The default is 3.
-        vars_to_send: default: vars_to_send = ['Intensity, Interferogram', 'Coherence']
+    window : ``int``
+        Integer averaging window size. The default is 3.
+    vars_to_send:  list
+        default: vars_to_send = ['Intensity, Interferogram', 'Coherence']
         can in addition take among: 'IntensityAvgComplexMasterSlave',
         'IntensityAvgMaster', 'IntensityAvgSlave'
 
@@ -266,17 +269,11 @@ def compute_multilooking_Master_Slave(ds, dict_L1B_process):
     Exception
         Raises exception if `SigmaImageSingleLookRealPart` is not a 2D variable
     """
-    
-    default_window=3        # Default value for the rolling mean
     list_vars_to_send = set(['Intensity', 'Interferogram', 'Coherence',
                              'IntensityAvgComplexMasterSlave',
                             'IntensityAvgMaster', 'IntensityAvgSlave'])
-
-    vars_to_send = dict_L1B_process.get('vars_to_send', 
-                                        ['Intensity', 'Interferogram', 'Coherence'])
-    window = dict_L1B_process.get('window', default_window)
-    
-    if len(set(vars_to_send).difference(list_vars_to_send)) > 0:
+    vars_to_send = set(vars_to_send)
+    if len(vars_to_send.difference(list_vars_to_send)) > 0:
         raise Exception(f"vars_to_send should be within the following variables: {list_vars_to_send}")
 
     if len(ds.SigmaImageSingleLookRealPart.dims) > 2:
@@ -761,7 +758,7 @@ def processing_OSCAR_L1AP_to_L1B(L1AP_folder, campaign, acq_date, track, dict_L1
         logger.info(f"Begining of the processing of {L1AP_file_names[i]}")
         ds_dict[antenna] = seastar.oscar.level1.replace_dummy_values(
                 ds_dict[antenna], dummy_val=float(ds_dict[antenna].Dummy.data))
-        ds_ml[antenna] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds_dict[antenna], dict_L1B_process)
+        ds_ml[antenna] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds_dict[antenna], dict_L1B_process['window'])
         ds_ml[antenna]['Polarization'] = seastar.oscar.level1.check_antenna_polarization(ds_dict[antenna])
         ds_ml[antenna]['AntennaAzimuthImage'] = seastar.oscar.level1.compute_antenna_azimuth_direction(ds_dict[antenna], antenna=antenna)
         ds_ml[antenna]['TimeLag'] = seastar.oscar.level1.compute_time_lag_Master_Slave(ds_dict[antenna], options='from_SAR_time')         # Time difference between Master and Slave
@@ -791,10 +788,13 @@ def processing_OSCAR_L1AP_to_L1B(L1AP_folder, campaign, acq_date, track, dict_L1
     # Checking dataset attributes
     ds_L1B = seastar.oscar.tools.check_attrs_dataset(ds_L1B)
 
+    #Updating of the CodeVersion in the attrs:
+    ds_L1B.attrs["CodeVersion"] = __version__
+
     # Updating of the history in the attrs:
     current_history = ds_L1B.attrs.get("History", "")                                               # Get the current history or initialize it
-    new_entry = f"{dt.now(timezone.utc).strftime("%d-%b-%Y %H:%M:%S")} L1B processing."            # Create a new history entry
-    updated_history = f"{current_history}\n{new_entry}" if current_history else new_entry       # Append to the history
+    new_entry = f"{dt.now(timezone.utc).strftime("%d-%b-%Y %H:%M:%S")} L1B processing."             # Create a new history entry
+    updated_history = f"{current_history}\n{new_entry}" if current_history else new_entry           # Append to the history
     ds_L1B.attrs["History"] = updated_history                                                       # Update the dataset attributes
 
     # Defining filename for datafile
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -677,7 +677,7 @@ def track_title_to_datetime(start_time):
     return np.datetime64(dt.strptime(start_time, '%Y%m%dT%H%M'))
 
 
-def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=False):
+def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, write_nc=False):
     """
     Processing chain from L1AP tp L1B.
     
@@ -685,14 +685,15 @@ def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=Fa
     
     Parameters
     ----------
-        L1AP_path : ``str``
-            Path of the L1AP files.
+        L1AP_folder : ``str``
+            Folder of the L1AP files.
         campaign : ``str``
-            Campaign name.
+            Campaign name as defined in config/Campaign_name_lookup.ini file. Example of campaign name: 202205_IroiseSea, 202305_MedSea.
         acq_date : ``str``
-            Date of the data acquisition.
+            Date of the data acquisition with the format "YYYYMMDD".
         track : ``str``
-            Track name.
+            Track name as defined in config/XXX_TrackNames.ini file. Format should be "Track_x" for tracks over ocean and 
+            "Track_Lx" for tracks over land with "x" the number of the track.
         
         write_nc (bool, optional): 
             Argument to write the data in a netcdf file. Defaults to False.
@@ -702,6 +703,9 @@ def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=Fa
         ds_L1B: ``xr.Dataset``
             Xarray dataset of the L1B OSCAR data.
     """
+
+    # Checking campaign, acquisition date and track name
+    # TODO: add a chek for campaign, acquisition date and track name
 
     # Getting the date for every tracks in the dict.
     track_names_dict = seastar.utils.readers.read_OSCAR_track_names_config(campaign, acq_date)
@@ -714,7 +718,7 @@ def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=Fa
         logger.warning(f"Track '{track}' not found in track_names_dict. The code will crash.")
         
     # Loading the file names of the files corresponding to date_of_track (triplet of file - one per antenna)
-    L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_path + "/*" + date_of_track + "*.nc"))]
+    L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_folder + "/*" + date_of_track + "*.nc"))]
 
     #-----------------------------------------------------------
     #               L1B PROCESSING
@@ -734,38 +738,25 @@ def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=Fa
 
     logger.info(f"Opening track: {track} on day: {acq_date}")
 
-    ds = seastar.oscar.tools.load_L1AP_OSCAR_data(L1AP_path, L1AP_file_names) 
+    ds_dict = seastar.oscar.tools.load_L1AP_OSCAR_data(L1AP_folder, L1AP_file_names) 
 
     # Getting the antenna identificators
     antenna_ident = seastar.oscar.tools.identify_antenna_location_from_filename(L1AP_file_names)
     logger.info(f"Antenna: {antenna_ident}")
-    
-    log_message_nesz = "\n".join(["Noise Equivalent Sigma Zero:",
-                                    f"Mid: {ds[0].SigmaNoise.values}",
-                                    f"Fore: {ds[1].SigmaNoise.values}",
-                                    f"Aft:  {ds[2].SigmaNoise.values}"
-                                    ])
-    logger.info(log_message_nesz)
-
-    log_message_nesz_slave = "\n".join(["Noise Equivalent Sigma Zero Slave:",
-                                        f"Fore: {ds[1].SigmaNoiseSlave.values}",
-                                        f"Aft: {ds[2].SigmaNoiseSlave.values}"
-                                        ])
-    logger.info(log_message_nesz_slave)
 
     ds_ml = dict()
-    for i in list(ds.keys()):
+    for i in list(ds_dict.keys()):
         logger.info(f"Begining of the processing of {L1AP_file_names[i]}")
-        ds[i] = seastar.oscar.level1.replace_dummy_values(
-                ds[i], dummy_val=float(ds[i].Dummy.data))
-        ds_ml[i] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds[i], window)
-        ds_ml[i]['Polarization'] = seastar.oscar.level1.check_antenna_polarization(ds[i])
-        ds_ml[i]['AntennaAzimuthImage'] = seastar.oscar.level1.compute_antenna_azimuth_direction(ds[i], antenna=antenna_ident[list(ds.keys()).index(i)])
-        ds_ml[i]['TimeLag'] = seastar.oscar.level1.compute_time_lag_Master_Slave(ds[i], options='from_SAR_time')         # Time difference between Master and Slave
-        ds_ml[i][vars_to_keep] = ds[i][vars_to_keep]
+        ds_dict[i] = seastar.oscar.level1.replace_dummy_values(
+                ds_dict[i], dummy_val=float(ds_dict[i].Dummy.data))
+        ds_ml[i] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds_dict[i], window)
+        ds_ml[i]['Polarization'] = seastar.oscar.level1.check_antenna_polarization(ds_dict[i])
+        ds_ml[i]['AntennaAzimuthImage'] = seastar.oscar.level1.compute_antenna_azimuth_direction(ds_dict[i], antenna=antenna_ident[list(ds_dict.keys()).index(i)])
+        ds_ml[i]['TimeLag'] = seastar.oscar.level1.compute_time_lag_Master_Slave(ds_dict[i], options='from_SAR_time')         # Time difference between Master and Slave
+        ds_ml[i][vars_to_keep] = ds_dict[i][vars_to_keep]
         ds_ml[i]['RadialSurfaceVelocity'] = seastar.oscar.level1.compute_radial_surface_velocity(ds_ml[i])
         ds_ml[i]['TrackTime'] = seastar.oscar.level1.track_title_to_datetime(ds_ml[i].StartTime)
-        ds_ml[i]['Intensity_dB'] = 10*np.log10(ds_ml[i].Intensity)
+        ds_ml[i]['Intensity_dB'] = seastar.utils.tools.lin2dB(ds_ml[i].Intensity)
 
         #Rolling median to smooth out TimeLag errors
         if not np.isnan(ds_ml[i].TimeLag).all():
@@ -798,7 +789,7 @@ def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=Fa
 
     # Write the data in a NetCDF file
     if write_nc: 
-        path_new_data = os.path.join(os.path.dirname(L1AP_path).replace("L1AP", "L1B"), os.path.basename(L1AP_path))
+        path_new_data = os.path.join(os.path.dirname(L1AP_folder).replace("L1AP", "L1B"), os.path.basename(L1AP_folder))
         if not os.path.exists(path_new_data):
             os.makedirs(path_new_data, exist_ok=True)
             logger.info(f"Created directory {path_new_data}")
--- a/seastar/oscar/level1.py
+++ b/seastar/oscar/level1.py
@@ -40,10 +40,10 @@ def fill_missing_variables(ds_dict, antenna_list):
     fore_id = list(ds_dict.keys())[antenna_list.index('Fore')]
     mid_id = list(ds_dict.keys())[antenna_list.index('Mid')]
     aft_id = list(ds_dict.keys())[antenna_list.index('Aft')]
-    antenna_list = [fore_id, mid_id, aft_id]
+    antenna_list2 = [fore_id, mid_id, aft_id]
 
-    for antenna_1 in antenna_list:
-        for antenna_2 in [a for a in antenna_list if a not in [antenna_1]]:
+    for antenna_1 in antenna_list2:
+        for antenna_2 in [a for a in antenna_list2 if a not in [antenna_1]]:
             for var in ds_dict[antenna_1].data_vars:
                 if var not in ds_dict[antenna_2].data_vars:
                     ds_dict[antenna_2][var] = xr.DataArray(data=np.nan)
@@ -745,8 +745,8 @@ def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, write_nc=
     logger.info(f"Antenna: {antenna_list}")
 
     ds_ml = dict()
-    for antenna in antenna_list:
-        logger.info(f"Begining of the processing of {L1AP_file_names[antenna]}")
+    for i, antenna in enumerate(antenna_list):
+        logger.info(f"Begining of the processing of {L1AP_file_names[i]}")
         ds_dict[antenna] = seastar.oscar.level1.replace_dummy_values(
                 ds_dict[antenna], dummy_val=float(ds_dict[antenna].Dummy.data))
         ds_ml[antenna] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds_dict[antenna], window)
@@ -756,7 +756,7 @@ def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, write_nc=
         ds_ml[antenna][vars_to_keep] = ds_dict[antenna][vars_to_keep]
         ds_ml[antenna]['RadialSurfaceVelocity'] = seastar.oscar.level1.compute_radial_surface_velocity(ds_ml[antenna])
         ds_ml[antenna]['TrackTime'] = seastar.oscar.level1.track_title_to_datetime(ds_ml[antenna].StartTime)
-        ds_ml[antenna]['Intensity_dB'] = seastar.utils.tools.lin2dB(ds_ml[antenna].Intensity)
+        ds_ml[antenna]['Intensity_dB'] = seastar.utils.tools.lin2db(ds_ml[antenna].Intensity)
 
         #Rolling median to smooth out TimeLag errors
         if not np.isnan(ds_ml[antenna].TimeLag).all():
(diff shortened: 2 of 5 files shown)


seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+                                        ])
+    logger.info(log_message_nesz_slave)
+
+    ds_ml = dict()
+    for i in list(ds.keys()):
+        logger.info(f"Begining of the processing of {L1AP_file_names[i]}")
+        ds[i] = seastar.oscar.level1.replace_dummy_values(
+                ds[i], dummy_val=float(ds[i].Dummy.data))
+        ds_ml[i] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds[i], window)
+        ds_ml[i]['Polarization'] = seastar.oscar.level1.check_antenna_polarization(ds[i])
```
ACHMartin, line 762: the name is not appropriate, we don"t "check", we "get". Need to change the "units". To be raised as an issue (if agreed)

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+        ds_ml[i]['AntennaAzimuthImage'] = seastar.oscar.level1.compute_antenna_azimuth_direction(ds[i], antenna=antenna_ident[list(ds.keys()).index(i)])
+        ds_ml[i]['TimeLag'] = seastar.oscar.level1.compute_time_lag_Master_Slave(ds[i], options='from_SAR_time')         # Time difference between Master and Slave
```
ACHMartin, line 764: @DavidMcCann-NOC for "compute_time_lag_Master_Slave" is options ='from_aircraft_velocity' still an option?
Do we have "Baseline" as a field in the L1AP? Should we have it as an option for "processing_L1AP_to_L1B"?

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+        ds_ml[i][vars_to_keep] = ds[i][vars_to_keep]
+        ds_ml[i]['RadialSurfaceVelocity'] = seastar.oscar.level1.compute_radial_surface_velocity(ds_ml[i])
+        ds_ml[i]['TrackTime'] = seastar.oscar.level1.track_title_to_datetime(ds_ml[i].StartTime)
+        ds_ml[i]['Intensity_dB'] = 10*np.log10(ds_ml[i].Intensity)
```
ACHMartin, line 768: please use seastar.tools.lin2dB

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+
+        #Rolling median to smooth out TimeLag errors
```
ACHMartin, line 770: Is it needed? (@DavidMcCann-NOC do you remember why we had this?)

Anyway, we need this before line 766 (I believe). Logically it will go just after "compute_time_lag.."

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+    logger.info(f"Build L1 dataset for :  {track} of day: {acq_date}")
+    
+    ds_L1B = seastar.oscar.level1.merge_beams(ds_ml, antenna_ident)
+    del ds_ml   
+    ds_L1B = ds_L1B.drop_vars(['LatImage', 'LonImage'], errors='ignore')
+    
+    # Checking dataset attributes
+    ds_L1B = seastar.oscar.tools.check_attrs_dataset(ds_L1B)
+
+    # Updating of the history in the attrs:
```
ACHMartin, line 790: should we put in history what are the L1AP? + the options of the processing?

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+    current_history = ds_L1B.attrs.get("History", "")                                               # Get the current history or initialize it
+    new_entry = f"{dt.now(timezone.utc).strftime("%d-%b-%Y %H:%M:%S")} L1B processing."            # Create a new history entry
+    updated_history = f"{current_history}\n{new_entry}" if current_history else new_entry       # Append to the history
+    ds_L1B.attrs["History"] = updated_history                                                       # Update the dataset attributes
+
+    # Defining filename for datafile
+    ds_L1B, filename = seastar.oscar.tools.formatting_filename(ds_L1B)
```
ACHMartin, line 797: why "ds_L1B" is an output of "formatting_filename"? We don't touch it at all in the function?

seastar/utils/readers.py
```diff
@@ -158,3 +159,23 @@ def read_OSCAR_track_names_config(campaign, flight):
     track_names_config.read(os.path.join(os.path.join('config', config_file_name)))
     track_names_dict = dict(track_names_config.items(flight))
     return track_names_dict
+
+def read_campaign_config():
```
ACHMartin, line 163: what is the difference between read_config and read_campaign_config, should we have both?

Could we have the same function, but with different input if we want to read the campaign or the tracks?

ACHMartin (commented):
Sorry, many comments. Some relatively important comments questionning the functions API and functions existence. If too difficult to deal with shortly, I suggest to open new issues.

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+def processing_OSCAR_L1APtoL1B(L1AP_path, campaign, acq_date, track, write_nc=False):
+    """
+    Processing chain from L1AP tp L1B.
+    
+    This function processes the OSCAR data from L1AP to L1B. It needs a triplet of L1AP file to generate a L1B file with data of the three antennas.
+    
+    Parameters
+    ----------
+        L1AP_path : ``str``
+            Path of the L1AP files.
```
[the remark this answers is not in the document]
elemerle: This is the L1AP folder where the files are. I can change L1AP_path into L1AP_folder.

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+        campaign : ``str``
+            Campaign name.
+        acq_date : ``str``
+            Date of the data acquisition.
+        track : ``str``
+            Track name.
```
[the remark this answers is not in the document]
elemerle: The format of the Track is "Track_x" for tracks in ocean and "Track_Lx" for tracks over land. 
For some tracks during the Iroise Sea campaign the format is "Track_xb". 

I think it is easier to handle to not allow list of tracks. The user can choose to run the function for a list of Track.

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+    
+    # Getting the date for the track we are interested in
+    if track in track_names_dict.values():
+        date_of_track = [key for key, v in track_names_dict.items() if v == track][0]
+        logger.info(f"Track '{track}' found, acquisition time: {date_of_track}")
+    else:
+        logger.warning(f"Track '{track}' not found in track_names_dict. The code will crash.")
+        
+    # Loading the file names of the files corresponding to date_of_track (triplet of file - one per antenna)
+    L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_path + "/*" + date_of_track + "*.nc"))]
```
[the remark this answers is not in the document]
elemerle: Yes you are right, I assume that now @DavidMcCann-NOC improved the way to identify antenna in the filename we can keep the the path within the file name. However, if the user have a path with 2numbers in a row the antenna ident function may crash. That is why I thought that it was better to have the path and the filename in another variable.

To avoid the separator we can use os.join:
L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(os.path.join(L1AP_path, f"*{date_of_track}*.nc"))]

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+            'CentralFreq',
+            'OrbitHeadingImage'
+            ]
+
+    logger.info(f"Opening track: {track} on day: {acq_date}")
+
+    ds = seastar.oscar.tools.load_L1AP_OSCAR_data(L1AP_path, L1AP_file_names) 
+
+    # Getting the antenna identificators
+    antenna_ident = seastar.oscar.tools.identify_antenna_location_from_filename(L1AP_file_names)
```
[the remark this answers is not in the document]
elemerle: I think this: "I suggest we do the antenna identification within "load_L1AP_OSCAR_data", the output of this function would be a dict_ds with keys 'Fore', 'Mid', 'Aft'." is a good idea. @DavidMcCann-NOC do you agree?

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+    logger.info(f"Antenna: {antenna_ident}")
+    
+    log_message_nesz = "\n".join(["Noise Equivalent Sigma Zero:",
+                                    f"Mid: {ds[0].SigmaNoise.values}",
```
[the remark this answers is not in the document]
elemerle: We are not sure but we put all the chances on our side when we sort the file but I agree this is not robust. The idea with dict_ds with the antenna as key is nice

seastar/oscar/level1.py
elemerle: I guess this is a check and get:
    Check polarization and standardize format.

    Check the Tx and Rx polarization variables in an OSCAR dataset and
    create a new 'Polarization' variable in a standard 'HH', 'VV' style
    format.
    
 We can raise an issue to adress this

seastar/oscar/level1.py
elemerle: We can open an issue to discuss what we want to put in the history.

seastar/oscar/level1.py
elemerle: You are right this is an omission from the previous version where I checked the attrs inside this function. I will delete it

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+            'LatImage',
+            'LonImage',
+            'IncidenceAngleImage',
+            'LookDirection',
+            'SquintImage',
+            'CentralFreq',
+            'OrbitHeadingImage'
+            ]
+
+    logger.info(f"Opening track: {track} on day: {acq_date}")
```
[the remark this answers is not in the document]
elemerle: This has been adressed in the last commit

seastar/oscar/level1.py
elemerle: done

seastar/oscar/level1.py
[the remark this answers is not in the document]
elemerle: I have use the os.join instead of "/"

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+
+    ds = seastar.oscar.tools.load_L1AP_OSCAR_data(L1AP_path, L1AP_file_names)
```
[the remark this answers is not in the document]
elemerle: ds was renammed in ds_dict with antenna as keys

seastar/oscar/tools.py
```diff
@@ -30,15 +30,16 @@ def load_L1AP_OSCAR_data(file_path, file_list):
...
+        keys as the name of each antenna as a ``str``
 
     """
-    ds = dict()
+    ds_dict = dict()
     for ind, file_name in enumerate(file_list):
-        ds[ind] = readNetCDFFile(os.path.join(file_path,file_name))
-    return ds
+        antenna = identify_antenna_location_from_filename(file_name)
+        ds_dict[antenna] = readNetCDFFile(os.path.join(file_path,file_name))
```
ACHMartin, line 41: so much easier this way :) Thanks

seastar/oscar/tools.py
```diff
@@ -125,18 +126,18 @@ def antenna_idents(ds):
...
+    antenna_list = list()
     for i in list(ds.keys()):
-        antenna_id.append(identify_antenna_location(ds[i]))
-    return antenna_id
+        antenna_list.append(identify_antenna_location(ds[i]))
+    return antenna_list
 
 
-def identify_antenna_location_from_filename(file_list):
+def identify_antenna_location_from_filename(file_name):
```
ACHMartin, line 140: @DavidMcCann-NOC, please check if you're happy with this change in the API.

If necessary, I think we can keep the previous API with the file_list. At the beginning of the function, we could test if it is a list or a string, if a string, we change it as a single element list.

seastar/oscar/tools.py
```diff
@@ -147,19 +148,16 @@ def identify_antenna_location_from_filename(file_list):
...
     Returns
     -------
-    antenna_id : ``list``
-        List of antenna identifiers as strings like ['Mid', 'Fore', 'Aft']
+    antenna : ``str``
+        Name of antenna as strings like 'Mid', 'Fore' or 'Aft'.
 
     """
     r = re.compile(r'(?:\D*\d\D*\d\D*)$')
-    antenna_id = list()
```
ACHMartin, line 155: @DavidMcCann-NOC, it would be nice if you could explain what you're doing with the Regular Expression with the "identify_antenna_location_from_filename" function on line 154. I think I vaguely understand what it is doing. But certainly better if it is spelled out. => issue?

seastar/oscar/tools.py
ACHMartin: @elemerle, for a big change of API like this. You should highlight it, because it can cause bugs which are difficult to see. While reviewing the code, it is easy to mis that it change the API.

seastar/oscar/level1.py
```diff
@@ -26,7 +26,7 @@ def fill_missing_variables(ds_dict, antenna_id):
     ds : ``dict``
         OSCAR data stored as a dict with antenna number as keys and loaded
         data in ``xarray.Dataset`` format as values
-    antenna_ident : ``list``
+    antenna_list : ``list``
```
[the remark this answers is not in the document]
elemerle: Indeed, I didn't check properly this function, lines 40-43 can be deleted

seastar/oscar/level1.py
```diff
@@ -82,7 +82,7 @@ def merge_beams(ds_dict, antenna_id):
     ds_level1 = xr.concat(list(ds_dict.values()),
                           'Antenna', join='outer',
                           coords='all')
-    ds_level1 = ds_level1.assign_coords(Antenna=('Antenna', antenna_id))
+    ds_level1 = ds_level1.assign_coords(Antenna=('Antenna', antenna_list))
```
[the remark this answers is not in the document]
elemerle: Yes list(ds_dict.keys()) return ['Mid', 'Fore', 'Aft'] in this order. So key_list[0]='Mid' etc

seastar/oscar/level1.py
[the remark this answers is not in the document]
elemerle: I think this function should be revised because antenna_1, antenna_2 is not clear to me. Maybe we can raise an issue for improving this function

DavidMcCann-NOC:
> def get_base_directory():
> """Return the base directory based on the operating system."""
> if platform.system() == "Windows":
> return "N:\"
> elif platform.system() == "Linux" :
> return "/PROJETS/"
> else:
> logger.warning(f"Unknown system name: {platform.system()}. Your path to data may be corrupted.")
> return None

There is a means in the repo to log this base path in a config file on your system, e.g. I used to use:
`file_paths = ss.utils.readers._read_config(config_file='config_LIVPC141.ini')`
(with config_LIVPC141.ini being my own local version of config.ini). That way you can set the local machine name (`platform.node()`) and have it pull up the base directory from there every time. I found it useful to keep track of multiple machines / OS i was using on my branch.

seastar/oscar/tools.py
DavidMcCann-NOC: Made an issue

seastar/oscar/level1.py
DavidMcCann-NOC: > Is it needed? (@DavidMcCann-NOC do you remember why we had this?)
> 
> Anyway, we need this before line 766 (I believe). Logically it will go just after "compute_time_lag.."

The smoothing of time lag data? This was introduced because there used to be some very intense lines in the time lag imagery that could be filtered out with a rolling median filter window. We'll have to check if they are still present in the latest data version - they used to cause absolute havoc with the RSV

seastar/oscar/tools.py
DavidMcCann-NOC: I mean, in a very real way these changes have effectively broken my old L1B processing chain completely, but thats not necessarily a bad thing! I'll give it a test and see

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+    # Loading the file names of the files corresponding to date_of_track (triplet of file - one per antenna)
+    L1AP_file_names = [os.path.basename(file) for file in sorted(glob.glob(L1AP_path + "/*" + date_of_track + "*.nc"))]
+
+    #-----------------------------------------------------------
+    #               L1B PROCESSING
+    #-----------------------------------------------------------
+    ds_L1B = dict()
+    window = 7              # window can be 3 or 7
+
+    vars_to_keep = [
```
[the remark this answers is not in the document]
DavidMcCann-NOC: Agreed - some times it is useful to change the vars that get kept or add more in, but these are definitely defaults

DavidMcCann-NOC:
@elemerle - make sure to update `_version.py`, i guess we're at 2025.04.1 here?

seastar/oscar/level1.py
```diff
@@ -670,3 +675,127 @@ def track_title_to_datetime(start_time):
...
+    new_entry = f"{dt.now(timezone.utc).strftime("%d-%b-%Y %H:%M:%S")} L1B processing."            # Create a new history entry
+    updated_history = f"{current_history}\n{new_entry}" if current_history else new_entry       # Append to the history
+    ds_L1B.attrs["History"] = updated_history                                                       # Update the dataset attributes
+
+    # Defining filename for datafile
+    filename = seastar.oscar.tools.formatting_filename(ds_L1B)
+
+    # Write the data in a NetCDF file
+    if write_nc: 
+        path_new_data = os.path.join(os.path.dirname(L1AP_folder).replace("L1AP", "L1B"), os.path.basename(L1AP_folder))
```
DavidMcCann-NOC, line 792: very nice use of the os functions!

seastar/oscar/level1.py
```diff
@@ -670,3 +675,127 @@ def track_title_to_datetime(start_time):
         Dataset start time in the form "YYYYMMDDTHHMM"
     """
     return np.datetime64(dt.strptime(start_time, '%Y%m%dT%H%M'))
+
+
+def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, write_nc=False):
```
DavidMcCann-NOC, line 680: Really good to have this all start to be properly formalised in a function, thanks for that Eva!

I'd prefer some more '_' in the name, e.g. `processing_OSCAR_L1AP_to_L1B()` for readability

seastar/oscar/level1.py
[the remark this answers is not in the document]
ACHMartin: > I think this function should be revised because antenna_1, antenna_2 is not clear to me. Maybe we can raise an issue for improving this function

The loop could be explained. The idea is to check the existence of variables for each antennas dataset. If it does not exist, we put a variable with NaN.
What happens before for identifying the antennas, looks a bit complicated to me.
Please check and test this function before deleting. @DavidMcCann-NOC is using it in the processing chain. Without it, we had big issues.

DavidMcCann-NOC:
On testing, I get the following error (with traceback):

```
usage: ipykernel_launcher.py [-h] [-v]
ipykernel_launcher.py: error: unrecognized arguments: -f C:\Users\dmca\AppData\Roaming\jupyter\runtime\kernel-1471f7d7-d0b3-4e70-8c6d-7c3a46786b6b.json

An exception has occurred, use %tb to see the full traceback.

SystemExit: 2

---------------------------------------------------------------------------
SystemExit                                Traceback (most recent call last)
Cell In[2], line 4
      2 import numpy as np
      3 import scipy as sp
----> 4 import seastar as ss
      5 import warnings
      6 import os

File C:\projects\python\seastar_project\seastar\__init__.py:1
----> 1 import seastar.utils
      2 import seastar.retrieval
      3 import seastar.oscar

File C:\projects\python\seastar_project\seastar\utils\__init__.py:2
      1 import seastar.utils.tools
----> 2 import seastar.utils.readers

File C:\projects\python\seastar_project\seastar\utils\readers.py:10
      7 from configparser import ConfigParser
      8 import hashlib
---> 10 from _logger import logger
     12 def _read_config(config_file='config.ini'):
     13     """
     14     Read configuration file.
     15 
   (...)
     35 
     36     """

File C:\projects\python\seastar_project\_logger.py:7
      5 parser = argparse.ArgumentParser(description="Run the script with optional verbose mode.")
      6 parser.add_argument("-v", "--verbose", action="store_true", help="Enable verbose mode (INFO logs)")
----> 7 args = parser.parse_args()
      9 # Logger configuration
     10 logging.basicConfig(
     11     level=logging.INFO if args.verbose else logging.WARNING,  # Set level based on --verbose
     12     format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
   (...)
     16     ]
     17 )

File ~\AppData\Local\anaconda3\envs\seastar\Lib\argparse.py:1893, in ArgumentParser.parse_args(self, args, namespace)
   1891 msg = _('unrecognized arguments: %s') % ' '.join(argv)
   1892 if self.exit_on_error:
-> 1893     self.error(msg)
   1894 else:
   1895     raise ArgumentError(None, msg)

File ~\AppData\Local\anaconda3\envs\seastar\Lib\argparse.py:2645, in ArgumentParser.error(self, message)
   2643 self.print_usage(_sys.stderr)
   2644 args = {'prog': self.prog, 'message': message}
-> 2645 self.exit(2, _('%(prog)s: error: %(message)s\n') % args)

File ~\AppData\Local\anaconda3\envs\seastar\Lib\argparse.py:2632, in ArgumentParser.exit(self, status, message)
   2630 if message:
   2631     self._print_message(message, _sys.stderr)
-> 2632 _sys.exit(status)

SystemExit: 2
```

My kernel appears to be having a serious error trying to use the `_logger` and can't `import seastar as ss` due to the new logging function being built into `ss.utils.readers()`. I'll have a look, but there might be a fundamental difference to how our two kernels have been set up. Are you using pip or conda to set up your environment? If the former, then we are using different methods which may be the source of differences in the kernel. I have no clue why though!

DavidMcCann-NOC:
> My kernel appears to be having a serious error trying to use the `_logger` and can't `import seastar as ss` due to the new logging function being built into `ss.utils.readers()`. I'll have a look, but there might be a fundamental difference to how our two kernels have been set up. Are you using pip or conda to set up your environment? If the former, then we are using different methods which may be the source of differences in the kernel. I have no clue why though!

Ah! Its because `args = parser.parse_args()` does not play well with Jupyter notebook. There is a fix - I can push it as a commit on this branch if you like?

seastar/oscar/level1.py
```diff
@@ -745,8 +745,8 @@ def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, write_nc=
     logger.info(f"Antenna: {antenna_list}")
 
     ds_ml = dict()
-    for antenna in antenna_list:
-        logger.info(f"Begining of the processing of {L1AP_file_names[antenna]}")
+    for i, antenna in enumerate(antenna_list):
+        logger.info(f"Begining of the processing of {L1AP_file_names[i]}")
```
ACHMartin, line 749: are we sure "i" refer to the same antenna?
It would be safer to read directly the filename in the attribute:
something like: "ds_dict[antenna].attrs.L1AP_filename"

seastar/oscar/level1.py
ACHMartin: Do you agree to have this just after "compute_time_lag" and in parallel we raise an issue to look after this (I guess it is low priority)

_logger.py
```diff
@@ -1,8 +1,14 @@
 import logging
+import argparse
+
+# Parse command-line arguments
+parser = argparse.ArgumentParser(description="Run the script with optional verbose mode.")
+parser.add_argument("-v", "--verbose", action="store_true", help="Enable verbose mode (INFO logs)")
+args = parser.parse_args()
```
DavidMcCann-NOC, line 7: Change to:
`args, unknown = parser.parse_known_args()`
To avoid bug with Jupyter notebook. Let me know if this doesn't work for you implementation and I'll figure out a different fix

seastar/oscar/level1.py
```diff
@@ -670,3 +675,127 @@ def track_title_to_datetime(start_time):
...
+        ds_dict[antenna] = seastar.oscar.level1.replace_dummy_values(
+                ds_dict[antenna], dummy_val=float(ds_dict[antenna].Dummy.data))
+        ds_ml[antenna] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds_dict[antenna], window)
+        ds_ml[antenna]['Polarization'] = seastar.oscar.level1.check_antenna_polarization(ds_dict[antenna])
+        ds_ml[antenna]['AntennaAzimuthImage'] = seastar.oscar.level1.compute_antenna_azimuth_direction(ds_dict[antenna], antenna=antenna)
+        ds_ml[antenna]['TimeLag'] = seastar.oscar.level1.compute_time_lag_Master_Slave(ds_dict[antenna], options='from_SAR_time')         # Time difference between Master and Slave
+        ds_ml[antenna][vars_to_keep] = ds_dict[antenna][vars_to_keep]
+        ds_ml[antenna]['RadialSurfaceVelocity'] = seastar.oscar.level1.compute_radial_surface_velocity(ds_ml[antenna])
```
DavidMcCann-NOC, line 757: Can we do this at L1B? Because the Interferograms haven't been calibrated yet at this point, so the RSV will be an uncalibrated version. I think this needs to be left until L1C, thoughts?

seastar/oscar/level1.py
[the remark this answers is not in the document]
DavidMcCann-NOC: > > I think this function should be revised because antenna_1, antenna_2 is not clear to me. Maybe we can raise an issue for improving this function
> 
> The loop could be explained. The idea is to check the existence of variables for each antennas dataset. If it does not exist, we put a variable with NaN. What happens before for identifying the antennas, looks a bit complicated to me. Please check and test this function before deleting. @DavidMcCann-NOC is using it in the processing chain. Without it, we had big issues.

I'm getting a little lost what we're taking about in this thread as its getting a bit long and unweildy, but yes - its vital to check for missing variables and add in NaNs as the Mid antenna files are smaller than the squinted ones, and there is no way to merge the beams together into one xarray dataset unless everything is standardised beforehand

seastar/oscar/level1.py
elemerle: Thanks Dave, Okay I'll modify the name with more "_".

seastar/oscar/level1.py
elemerle: I don't think the L1AP_filename appears in the attrs does it?
I guess this loop is safe because the antenna_list is written fromo the ds_dict.keys(): antenna_list = list(ds_dict.keys())

_logger.py
elemerle: Just tested it, this modification works for me I push it

seastar/oscar/level1.py
ACHMartin: I don't know, but it might be useful to add it?
Potentially only for the dataset (not necessary to be saved in the netCDF). It might be handy for the history as well.

seastar/oscar/level1.py
DavidMcCann-NOC: > Do you agree to have this just after "compute_time_lag" and in parallel we raise an issue to look after this (I guess it is low priority)

After our discussion just now i've raised an issue (#342 ) to make a new `TimeLag` filtering function as its absolutely still required even in the latest datasets. It should come just after `compute_time_lag_Master_Slave` and if we are to keep in `compute_radial_surface_velocity` it should come after the `TimeLag` filter.

seastar/oscar/level1.py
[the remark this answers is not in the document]
elemerle: Resolved in commit [1b62113]

seastar/oscar/level1.py
```diff
@@ -670,3 +675,136 @@ def track_title_to_datetime(start_time):
...
+                                        f"Aft: {ds[2].SigmaNoiseSlave.values}"
+                                        ])
+    logger.info(log_message_nesz_slave)
+
+    ds_ml = dict()
+    for i in list(ds.keys()):
+        logger.info(f"Begining of the processing of {L1AP_file_names[i]}")
+        ds[i] = seastar.oscar.level1.replace_dummy_values(
+                ds[i], dummy_val=float(ds[i].Dummy.data))
+        ds_ml[i] = seastar.oscar.level1.compute_multilooking_Master_Slave(ds[i], window)
```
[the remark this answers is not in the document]
elemerle: This is kept optionnal field with default value in the dict cf commit [1b62113]

seastar/utils/readers.py
elemerle: So should we keep the _read_config() appart and have one unique config reader for campaigns and Tracks name reader?

DavidMcCann-NOC: > So should we keep the _read_config() appart and have one unique config reader for campaigns and Tracks name reader?

For now, yes - we can plan changes to how the configs are handled in a separate piece of work. It would be good to make it more explicit (e.g., a file_path_config or something) and stored in the /config folder. But either way its an optional config for local path definition so not everyone will want or need it, but its good to have I think.

seastar/oscar/level1.py
```diff
@@ -724,17 +732,9 @@ def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, write_nc=
...
-    vars_to_keep = [
-            'LatImage',
-            'LonImage',
-            'IncidenceAngleImage',
-            'LookDirection',
-            'SquintImage',
-            'CentralFreq',
-            'OrbitHeadingImage'
-            ]
+
```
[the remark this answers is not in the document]
elemerle: ds_L1B is used line 774I I move the definition of L1B to line 772

seastar/oscar/level1.py
elemerle: The filter function is just after the timelag computation. The RSV calculation comes after the timelag filter

seastar/oscar/level1.py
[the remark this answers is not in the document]
ACHMartin: Ok, but you don't need to create it with `ds_L1B = dict()` as it is generated with `ds_L1B = seastar.oscar.level1.merge_beams(ds_ml, antenna_list)`

seastar/oscar/level1.py
DavidMcCann-NOC: > The filter function is just after the timelag computation. The RSV calculation comes after the timelag filter

I think I'm starting to get confused with all the comments, so perhaps we can look at this after its merged, but on my screen the RSV computation on `Line 757` needs to be moved to, e.g., `Line 766`, or `Lines 761-765` need to be moved to Line `756`. I.e., Compute TimeLag --> Median Filter --> RSV computation

seastar/oscar/level1.py
```diff
@@ -753,24 +752,25 @@ def processing_OSCAR_L1APtoL1B(L1AP_folder, campaign, acq_date, track, dict_L1B_
...
+        
+        
     ds_ml = seastar.oscar.level1.fill_missing_variables(ds_ml, antenna_list)
     
     #-----------------------------------------------------------
     
     # Building L1 dataset
     logger.info(f"Build L1 dataset for :  {track} of day: {acq_date}")
-    
+    ds_L1B = dict()
```
ACHMartin, line 773: you don't need to initialize `ds_L1B`

seastar/oscar/tools.py
```diff
@@ -224,7 +224,7 @@ def formatting_filename(ds):
         ds.attrs.get("Platform"),
         ds.attrs.get("ProcessingLevel"),
         ds.attrs.get("Track"),
-        f"{int(ds.attrs.get("MultiLookCrossRangeEffectiveResolution")).zfill(3)}x{int(ds.attrs.get("MultiLookGroundRangeEffectiveResolution")).zfill(3)}m",
+        f"{str(int(ds.attrs.get("MultiLookCrossRangeEffectiveResolution"))).zfill(3)}x{str(int(ds.attrs.get("MultiLookGroundRangeEffectiveResolution"))).zfill(3)}m",
```
ACHMartin, line 227: what about?
```
"{crossRes:03d}x{groundRes:03d}".format(
     crossRes=ds.attrs.get("MultiLookCrossRangeEffectiveResolution"),
     groundRes=ds.attrs.get("MultiLookGroundRangeEffectiveResolution")
)
```

seastar/oscar/tools.py
```diff
@@ -224,7 +224,7 @@ def formatting_filename(ds):
...
         ds.attrs.get("L2Processor", ""),  # Only for L2
```
ACHMartin, line 228: I didn't notice this before.
If the attributes is filled earlier (for whatever reason), the filename will be affected. Should you put an "if" on the product level? I remember you did something like this somewhere.

We should use "check_attrs_dataset" at the beginning of the function to check that all expected attributes exist.

ACHMartin (commented):
we are getting there :)

seastar/oscar/level1.py
```diff
@@ -712,8 +712,20 @@ def processing_OSCAR_L1AP_to_L1B(L1AP_folder, campaign, acq_date, track, dict_L1
...
-    # Checking campaign, acquisition date and track name
-    # TODO: add a chek for campaign, acquisition date and track name
+    # checking acq_date format:
+    if seastar.oscar.tools.is_valid_acq_date(acq_date):
+        logger.info("'acq_date' format is okay")
+    else: 
+        logger.error("'acq_date' should be a valid date string in 'YYYYMMDD' format ")
+        raise ValueError("'acq_date' should be a valid date string in 'YYYYMMDD' format ")
+
+    # Checking campaign name:
```
ACHMartin, line 722: To add:
`valid_campaign = ["202205_IroiseSea", "202305_MedSea"]`

seastar/oscar/tools.py
elemerle: It is done in processing chain:
    # Checking dataset attributes
    ds_L1B = seastar.oscar.tools.check_attrs_dataset(ds_L1B)

    # Updating of the history in the attrs:
    current_history = ds_L1B.attrs.get("History", "")                                               # Get the current history or initialize it
    new_entry = f"{dt.now(timezone.utc).strftime("%d-%b-%Y %H:%M:%S")} L1B processing."            # Create a new history entry
    updated_history = f"{current_history}\n{new_entry}" if current_history else new_entry       # Append to the history
    ds_L1B.attrs["History"] = updated_history                                                       # Update the dataset attributes

    # Defining filename for datafile
    filename = seastar.oscar.tools.formatting_filename(ds_L1B)

So do you want check_attrs in both processing chain and formatting_filename? or do I move it to formatting_filename?

ACHMartin: I don't know, but while reading the commit, I was wondering it could be a source of error, so I guess it is good to have it here.
I let you decide if it is needed in the processing chain.

seastar/oscar/level1.py
elemerle: cf issue #347

elemerle:
I think the branch is ready to be merged

ACHMartin (approved):

PR #345, opened 2025-04-03
Update L1A_to_L1AP_processing.m

DavidMcCann-NOC:
New function to compute grid resolutions directly from the CrossRange and GroundRange L1A data arrays, build a resolution string for the L1AP attributes and inject it into the NetCDF file.

### DIFF
--- a/matlab/metasensing/L1A_to_L1AP_processing.m
+++ b/matlab/metasensing/L1A_to_L1AP_processing.m
@@ -39,7 +39,6 @@ for file = 1 : num_L1A_files
     add_squint_to_L1AP_netcdf(L1AP_file_path, L1AP_file_name);
 
     add_orbit_images_to_L1AP_netcdf(L1AP_file_path, L1AP_file_name)
-    pause
 end
 
 end
--- a/matlab/metasensing/L1A_to_L1AP_processing.m
+++ b/matlab/metasensing/L1A_to_L1AP_processing.m
@@ -39,7 +39,7 @@ for file = 1 : num_L1A_files
     add_squint_to_L1AP_netcdf(L1AP_file_path, L1AP_file_name);
 
     add_orbit_images_to_L1AP_netcdf(L1AP_file_path, L1AP_file_name)
-
+    pause
 end
 
 end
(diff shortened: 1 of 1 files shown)


DavidMcCann-NOC:
I have included three attributes here (examples):
```
Resolution = 008x008m (for use as a resolution string)
CrossRangeResolution = 8.0 (grid resolution in the CrossRange direction in m)
GroundRangeResolution = 8.0 (grid resolution in the groundRange direction in m)
```
Happy to adapt these names to whatever you think best, or add in any other variations of attributes

ACHMartin:
> I have included three attributes here (examples):
> 
> ```
> Resolution = 008x008m (for use as a resolution string)
> CrossRangeResolution = 8.0 (grid resolution in the CrossRange direction in m)
> GroundRangeResolution = 8.0 (grid resolution in the groundRange direction in m)
> ```
> 
> Happy to adapt these names to whatever you think best, or add in any other variations of attributes

I would not use the "Resolution" at all. We can easily create it when needed,
For the L1AP, I suggest "SingleLookCrossRangeSpacing" and "SingleLookGroundRangeSpacing", we might drop the "Look" if you feel it is too long. If you don't like "Spacing", what about "GridResolution".

(FYI @elemerle ) For the L1B and L1C after multilooking, I suggest to have "MultiLookCrossRangeEffectiveResolution" = window * SingleLookCrossRangeSpacing, but given we are using rolling, we don't change the grid resolution, we can have "MultiLookCrossRangeSpacing".

As far as possible, I suggest to calculate from the data themselves (perhaps we could have a short function for that eg "calculate_grid_spacing(xr.Dataset:ds, str:dimension)")

ACHMartin (approved):

elemerle:
Okay I will add the "MultiLookCrossRangeSpacing" in the compute multilooking function

DavidMcCann-NOC:
> SingleLookCrossRange

Made these changes with [26c46a2](https://github.com/ACHMartin/seastar_project/pull/345/commits/26c46a2c6893db3a2623fb95c2205a147d1e8ba0)

PR #355, opened 2025-04-22
two new functions to generate environmental fields

ACHMartin:
@DavidMcCann-NOC, you will see I commented a large part of the function "generate_constant_env_field". Initially I did it specifically for wind and current, and in fact I believe it can be more generic. Hence this new proposition.
Please let me know what do you think, and I will remove or restore these commented code.
Thanks

### DIFF
--- a/seastar/performance/scene_generation.py
+++ b/seastar/performance/scene_generation.py
@@ -362,12 +362,10 @@ def satellite_looking_geometry(input):
     return sat_geometry
 
 
-def generate_constant_env_field(da: xr.DataArray, env: dict):
+def generate_constant_env_field(da: xr.DataArray, env: dict) -> xr.Dataset:
     '''
     Field generation of constant fields of the same dimension as the DataArray "da" with the "env" conditions
 
-    # WARNING by default take OceanSurfaceWind to create EarthRelativeWind
-    # #TODO do we need a function to check consistency between ERW, OSW and Current?
     Parameters
     ------------
     da : ``xarray.DataArray``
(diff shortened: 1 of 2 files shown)


seastar/utils/tools.py
```diff
@@ -146,6 +146,92 @@ def windUV2SpeedDir(u, v):
     return wspd, wdir
 
 
+def windCurrentComponent2to4(env: dict, basevarname: str):
+    '''
+    Complete U, V or Speed/Velocity, Direction
```
DavidMcCann-NOC, line 151: This could do with a better description I think?

seastar/performance/scene_generation.py
```diff
@@ -362,6 +362,80 @@ def satellite_looking_geometry(input):
     return sat_geometry
 
 
+def generate_constant_env_field(da: xr.DataArray, env: dict):
```
DavidMcCann-NOC, line 365: Just tested this - really useful function and much better than what we had previously

DavidMcCann-NOC, line 365: I notice that the variables in the function definition are in the form (var: type, var: type). This is the first time i've seen this, which isn't necessarily a bad thing

seastar/utils/tools.py
```diff
@@ -146,6 +146,92 @@ def windUV2SpeedDir(u, v):
     return wspd, wdir
 
 
+def windCurrentComponent2to4(env: dict, basevarname: str):
```
DavidMcCann-NOC, line 149: personally I prefer function names with _ delimiters between words to increase readability, although I know we have a few functions with this more compact naming style. For example this particular name doesn't elude to it's intended use very well, at least not to me.

DavidMcCann-NOC (approved):
Aside from a better function description for windCurrentComponent2to4 I think this is good to go

seastar/performance/scene_generation.py
ACHMartin: Indeed, I discovered this recently and I find it somehow useful:
https://docs.python.org/3/library/typing.html

seastar/utils/tools.py
ACHMartin: as discussed, the name of the function is changed to: "wind_current_component_conversion", with this new description:
Add U, V to Speed/Velocity, Direction or the other way around for "Current", 
    "Wind", "OceanSurfaceWind", "EarthRelativeWind".

PR #361, opened 2025-05-06
Improve performance: Optimize data loading

Rotsapy:


seastar/gmfs/nrcs.py
```diff
@@ -5,8 +5,51 @@
 import xarray as xr
 from scipy.interpolate import interpn
 import seastar
+from os.path import abspath, dirname, join
 
 
+def _load_gmf_table(fname, shape):
+    """Loads and reshapes a GMF table from a binary file."""
+    gmf_table = np.fromfile(fname, dtype=np.float32)
+    gmf_table = gmf_table[1:-1].reshape(shape, order="F")
```
ACHMartin, line 14: Could you please keep the comments from the original function, it will help to review it in future, particularly if we need to add new GMFs.

    # Remove head and tail
    # To access the table as a three-dimensional Fortran-ordered m x n x p matrix,
    # reshape it

seastar/gmfs/nrcs.py
```diff
@@ -5,8 +5,51 @@
...
+        gmf_model (str): The GMF model to load ('nscat4ds' or 'cmod7').
+
+    Returns:
+        tuple: A tuple containing the GMF table (numpy array) and the points
+               (tuple of numpy arrays for wspd, rdir, inci, and optionally pol).
+    """
+    dirpath = abspath(dirname(__file__))
+    m, n, p = 250, 73, 51  # Common dimensions
+
+    wspd = np.linspace(0.2, 50, m)
```
ACHMartin, line 31: same, please keep the comments

seastar/gmfs/nrcs.py
```diff
@@ -94,6 +137,8 @@ def compute_nrcs(L1_combined, aux_geo, gmf):
     nrcs.attrs['units'] = ['']
     return nrcs
 
+gmf_table_global_nscat4ds, points_global_nscat4ds = get_gmf_table_and_points('nscat4ds')
```
ACHMartin, line 140: Could you please explain how it works if the call of this function is out of a function?
I can see the variables are "called" within the nscat4ds function, but I don't understand how it works. In my mind, the variables should be undefined as never called...
Thanks in advance for the explanations.

seastar/gmfs/nrcs.py
```diff
@@ -5,8 +5,51 @@
...
 
 
+def _load_gmf_table(fname, shape):
+    """Loads and reshapes a GMF table from a binary file."""
+    gmf_table = np.fromfile(fname, dtype=np.float32)
+    gmf_table = gmf_table[1:-1].reshape(shape, order="F")
+    return gmf_table
+
+def get_gmf_table_and_points(gmf_model):
+    """
```
ACHMartin, line 18: The documentation is generated automatically using Sphinx, could you please use the same strict format as the other functions (for the ones which are up to date). Thanks

env/environment.yml
```diff
@@ -16,3 +16,4 @@ dependencies:
   - cartopy
   - utm
   - pytest
+  - tqdm
```
ACHMartin, line 19: should we add it to the "pyproject.toml" as well?

seastar/retrieval/level2.py
```diff
@@ -152,7 +151,10 @@ def run_find_minima(level1, noise, gmf, serial=False):
             lmoutmap = map(find_minima_parallel_task, input_mp)
         else:
             with multiprocessing.Pool() as pool:
-                lmoutmap = pool.map(find_minima_parallel_task, input_mp)
+                results = []
+                for result in tqdm(pool.imap(find_minima_parallel_task, input_mp), total=len(input_mp)):
+                    results.append(result)
```
ACHMartin, line 156: If we append the results as they come from the processors in parallel, we are not sure it will be ordered the same way as the input (which is critical for us).
I welcome a progress bar, but we need it to be ordered the same way. Should we initialize `results = [None] * len(input_mp)`? and then use the iteration values. It will not be as compact but will insure the order. Isn't it?

seastar/retrieval/cost_function.py
```diff
@@ -507,9 +508,6 @@ def smooth(x):
 #
 #
 # end
-
-    print("To Be Done find_initial_value")
```
ACHMartin, line 511: Please keep it, I know it is not pretty, but it is the only way for me to be sure to address this point. I know we should use GitHub issues, and we do, but this is critical (even if it lags 2 years behind...)

seastar/retrieval/cost_function.py
```diff
@@ -68,15 +70,14 @@ def fun_residual(variables, level1, noise, gmf):
     model_rsv_list = [None] * level1.Antenna.size
     for aa, ant in enumerate(level1.Antenna.data):
         # print(aa, ant)
-        model_rsv_list[aa] = seastar.gmfs.doppler\
-            .compute_total_surface_motion(level1.sel(Antenna=ant),
```
ACHMartin, line 72: @DavidMcCann-NOC, @elemerle , any view on this?
Personnaly, I tend to prefer to keep the full path.

seastar/retrieval/cost_function.py
```diff
@@ -158,7 +159,7 @@ def find_minima(level1_pixel, noise_pixel, gmf):
     for ii in [1,2,3]:
         try:
             lmout_dict = least_squares(
-                seastar.retrieval.cost_function.fun_residual,
```
ACHMartin, line 161: Ok, I agree this one doesn't make sense given it is in the same file.

ACHMartin (commented):
Thanks for the modifications. I have a critical comment concerning your modification for the multiprocessing with pool.map -> imap and in particular with the .append.

Could you please remind me what is the gain value to load the GMFs before hand? Did you try for the Doppler GMFs as well?
Thanks

seastar/retrieval/cost_function.py
elemerle: Both solutions have their pros and cons. But I would tend to agree with Javier (@Rotsapy) to write all the full path at the beginning so that you know that the information is in the import section.

seastar/retrieval/cost_function.py
elemerle: @Rotsapy you can use the logger if @ACHMartin is okay. So that you avoid printing and keep the info logged somewhere :)

ACHMartin: I will update it soon, no need to add the logger. Please just leave the print here. I will take care of it.

env/environment.yml
DavidMcCann-NOC: It will need adding to environments.yml in the /env folder as well if it ends up being a critical dependency, but we can cross that bridge when we need to

seastar/retrieval/level2.py
DavidMcCann-NOC: > If we append the results as they come from the processors in parallel, we are not sure it will be ordered the same way as the input (which is critical for us). I welcome a progress bar, but we need it to be ordered the same way. Should we initialize `results = [None] * len(input_mp)`? and then use the iteration values. It will not be as compact but will insure the order. Isn't it?

Agreed, we need to be careful about how the workers return the jobs. It might be good to include an option for the progressbar if we're going down that route too, in case we'd like to keep the execution silent and bypass the (likely small) overhead of the progressbar

Rotsapy: Issue opened to add option to toggle the progress bar #372

PR #371, opened 2025-06-11
Bugfixes for L2 operation on the servers

DavidMcCann-NOC:
Bug found in retrieval.level2

When running the retrieval on the server I got the following error:
```
File "/noc/mpoc/seastar/packages/seastar_project/seastar/retrieval/level2.py", line 413
    new_entry = f"{dt.now(timezone.utc).strftime("%d-%b-%Y %H:%M:%S")} L2 processing."             # Create a new history entry
                                                     ^
SyntaxError: invalid syntax
```

### DIFF
--- a/seastar/performance/scene_generation.py
+++ b/seastar/performance/scene_generation.py
@@ -469,7 +469,7 @@ def compute_truth_level1(
         inst: xr.Dataset, geo: xr.Dataset, gmf: dict, 
         write_nc: Optional[bool]=False, 
         main_path: Optional[str]='.'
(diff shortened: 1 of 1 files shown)


DavidMcCann-NOC:
Additional bugfixes required from other packages - all included here

DavidMcCann-NOC:
> That's funny that you and Adrien have the same error. Which version of python do you use?
> 
> It's okay for me you can merge.

Not sure, but I very recently installed a new version of the environment so I assume its the latest version

Issue #380, opened 2025-10-06
package organisation


ACHMartin:
We have the following structure (I only put the elements I want to discuss):

seastar
./examples <- do we need this module/function? Where does it come from?
./gmfs
./oscar
./oscar/level1
./oscar/tools
./performance
./retrieval
./retrieval/ambiguity_removal
./retrieval/cost_function
./retrieval/level2

The function "seastar.retrieval.level2.processing_OSCAR_L1_to_L2" is NOT within the "oscar" module, which is not intuitive.

Should we move the function within the "oscar" module? Where? Within a new "oscar/level2" module? In a renamed "oscar/level1" module? (oscar/processing_chain)

DavidMcCann-NOC:
I think a lot of the heritage of the dir structure is predicated on OSCAR being the only show in town, so I agree that OSCAR-specific code should be in the OSCAR folder ideally. I wouldn't have an issue with an OSCAR/retrieval folder or similar

PR #389, opened 2025-10-14
Cleaning seastar tree

ACHMartin:
I cleaned the seastar tree by moving all notebooks in a notebooks folder. @DavidMcCann-NOC, I kept all notebooks, but some might need to be cleaned. There is some notebooks which are nearly identical.

I removed some obsolete functions + config files.

It resolves some points raised in #380, but not all, yet.

### DIFF
--- a/seastar/utils/readers.py
+++ b/seastar/utils/readers.py
@@ -10,29 +10,6 @@ import hashlib
 from _logger import logger
 
 
-
-def readNetCDFFile(netCFD_path):
-    """
-    Read a netCDF file and returns it as an `xarray.Dataset`.
-
-    :param netCFD_path: path to the netCDF file
-    :type netCFD_path: ``str``
-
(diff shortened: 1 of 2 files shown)


seastar/utils/readers.py
```diff
@@ -9,89 +9,6 @@
 
 from _logger import logger
 
-def _read_config(config_file='config.ini'):
```
DavidMcCann-NOC: Is _read_config removed? If so, I do actually use this one

ACHMartin: indeed it is removed for now, but if needed I can put it back. Where is it used? @elemerle or @JWMichalski, do you used this function?

elemerle: No I don't use this function

DavidMcCann-NOC (approved):
Happy for this to go

Issue #401, opened 2025-10-22
Configparser not parsing track names configs in UNIX


DavidMcCann-NOC:
Upon running `ss.utils.readers.read_OSCAR_track_names_config(campaign, flight)` in UNIX it raises the following error:

```
Traceback (most recent call last):
  File "/noc/mpoc/seastar/environments/seastar/lib/python3.8/configparser.py", line 846, in items
    d.update(self._sections[section])
KeyError: '20230505'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test.py", line 6, in <module>
    flight_times = ss.utils.readers.read_OSCAR_track_names_config(campaign, flight)
  File "/noc/mpoc/seastar/packages/seastar_project/seastar/utils/readers.py", line 77, in read_OSCAR_track_names_config
    track_names_dict = dict(track_names_config.items(flight))
  File "/noc/mpoc/seastar/environments/seastar/lib/python3.8/configparser.py", line 849, in items
    raise NoSectionError(section)
configparser.NoSectionError: No section: '20230505'
```
Minimal replicatable example:

```
import seastar as ss

campaign = '202205_IroiseSea'
print(ss.utils.readers.read_OSCAR_track_names_config(campaign, '20220522'))

```
Also results in a KeyError for '20220522'

This is likely down to differences in end of line characters between Windows and UNIX, which will either need EOL characters specified in the config file (e.g., `/n`) or a change to the config reader to deal with UNIX systems

PR #402, opened 2025-10-23
Issue #401

DavidMcCann-NOC:
Fix to read in track names config in a platform agnostic manner and accounting for different install directories. navigates to seastar_project to make a root directory before navigating to /config. Tested working.

### DIFF
--- a/seastar/utils/readers.py
+++ b/seastar/utils/readers.py
@@ -74,12 +74,7 @@ def read_OSCAR_track_names_config(campaign, flight):
     logger.info(f"Reading Track name config file {config_file_name}...")
     track_names_config = ConfigParser()
     track_names_config.optionxform = str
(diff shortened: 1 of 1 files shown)


seastar/utils/readers.py
```diff
@@ -73,7 +74,8 @@ def read_OSCAR_track_names_config(campaign, flight):
     logger.info(f"Reading Track name config file {config_file_name}...")
     track_names_config = ConfigParser()
     track_names_config.optionxform = str
-    track_names_config.read(os.path.join(os.path.join('config', config_file_name)))
+    ROOT_DIR = find_seastar_project_root()
+    track_names_config.read(os.path.join(os.path.join(ROOT_DIR, 'config', config_file_name)))
```
ACHMartin: I believe there is one `join` extra.
`os.path.join(os.path.join(ROOT_DIR, 'config', config_file_name))` can be `os.path.join(ROOT_DIR, 'config', config_file_name)`

PR #414, opened 2026-01-28
Update L1A_to_L1AP_processing.m

DavidMcCann-NOC:
Error in config name.... apologies - I'm not sure how this tested working before, but I must have had a rogue variable sitting around in the matlab environment on the server. Evidently I didn't check....

This is the correct fix and is proven working on my test branch on the server right now.


ACHMartin (approved):
ok

PR #420, opened 2026-04-30
Updating phase sign convention configs and version for final delivery

DavidMcCann-NOC:


config/OSCAR_config.ini
```diff
@@ -7,10 +7,22 @@
...
+20220517 = -1
+20220522 = -1
+20220525 = -1
+20220526 = -1
+[phase_sign_convention_v20250704]
+20230505 = -1
+20230507 = -1
+[phase_sign_convention_v20260428]
+20220517 = 1
+[phase_sign_convention_v20260518]
```
ACHMartin, line 19: what is this version 2026 May 18

ACHMartin: should it be the 2026 April 28?

_version.py
```diff
@@ -7,5 +7,5 @@
 """
 
 
-__version__ = '2025.10.1'
+__version__ = '2026.04.1'
```
ACHMartin, line 10: Is it a major new version?
I am not sure. The only change is the sign convention. So I guess we could stay with 2025.10.1.
If you disagree, I let you take executive decision

ACHMartin (approved):

config/OSCAR_config.ini
DavidMcCann-NOC: The new data version is dated to the date of the FRM - I felt it would be cleaner that way. Slightly unorthodox for us to date to the future but this way it links nicely I think

_version.py
DavidMcCann-NOC: The dating convention is Year.Month.Sequential_Version so this is dated for now. Personally I think enough has changed for it to be updated, although I agree nothing major apart from the change in the sign convention config (which would make a huge difference processing the Iroise Sea using 2025.10.1).

PR #421, opened 2026-04-30
Update processing_chain.py

DavidMcCann-NOC:
Bug in processing chain for Netherlands data - raises an exception because the campaign isn't in the valid campaigns list.


ACHMartin (approved):
thnaks

PR #422, opened 2026-05-14
Delete poetry.lock

DavidMcCann-NOC:
Removal of poetry.lock file after security bulletins.

Poetry is the root of 13 high severity vulnerabilities in the repo [link](https://github.com/ACHMartin/seastar_project/security/dependabot), with some going back a number of years. As poetry provides non-essential services to the repository it is best to remove it and re-asses in the future.


ACHMartin (approved):
thanks

